PR #7, opened 2025-11-24
Genie 976/story/new allow to share chats with others

MayaCrmi:
# Chat Sharing Feature

This feature allows users to share their workflows via public chat links, allowing authenticated users to interact with shared workflows through a dedicated chat interface.

### 1. Workflow Sharing Toggle
- **UI Component**: Added `ShareWorkflow` component with a toggle switch to enable/disable public chat sharing
- **Visual Feedback**: 
  - Loading spinner and "Validating workflow..." message during enable process
  - Disabled state during validation to prevent multiple clicks
  - Clear error messages matching "Load Workflow" error format

image

image

image

### 2. Public Chat Link Generation
- **Format**: `/chat/{blueprint_id}` - Uses blueprint ID as the share token
- **Access Control**: Requires authentication (SSO) before accessing chat, auth_manager file changed to use the redirect URL argument conditionally depending on the referring URL
- **Link Display**: Shows generated share link with copy-to-clipboard functionality

image

### 3. Shared Chat Experience (`PublicChat`)
- **Dedicated Interface**: Full-screen chat experience for shared workflows
- **Header Components**:
  - UnifAI logo for brand consistency
  - Workflow name and owner information: `{workflowName} (workflow shared by {owner})`
  - Login/logout section matching main app authentication UI
- **Chat History Sidebar**:
  - User-specific chat history for the shared blueprint
  - Session selection with preview and timestamps
  - "New Chat" button to start fresh conversations
  - Delete chat functionality with confirmation modal

image

### 4. Chat Interface Modifications
- **Chat-Only Mode**: 
  - Hides agent thoughts/internal processing logs
  - Shows only a loading spinner during execution
  - Displays final answer after completion
- **Message Display**: 
  - User messages on the right (standard chat UI)
  - Bot messages on the left
  - Proper message role handling to prevent incorrect sender attribution

image

image

### 1. Sharing Endpoints (`/api/shares/`)
- **`/public-chat.enable`**: 
  - Ensures blueprint can be loaded, resolved, and compiled before enabling sharing
  - Sets `shared: true` on blueprint document
  - Returns share link upon success
- **`/public-chat.disable`**: 
  - Sets `shared: false` on blueprint document
  - Prevents new shared link access (in the /chat/{blueprint_id} URL)
- **`/public-chat.status.get`**: 
  - Returns current sharing status and share link
  - Used for UI state management
- **`/public-chat.validate`**: 
  - Validates share token (blueprint_id)
  - Returns blueprint information and ownership details to be displayed where necessary
  - Provides specific error messages for invalid/disabled workflows

### 2. Blueprint Document Structure
- **New Field**: `shared` (boolean) on blueprint documents
  - `true`: Workflow is publicly shareable via chat link
  - `false`: Sharing is disabled (default)

### 3. Session Creation
- **From Shared Links**: Sessions created from shared links include `fromSharedLink: true` metadata flag
- **Validation**: Same validation logic as regular session creation ensures only valid workflows can be shared

### 4. Blueprint Info Endpoint
- **`/api/blueprints/blueprint.info.get`**: 
  - Returns blueprint name and owner information
  - Used for displaying workflow details in shared chat interface

### 1. Redirect Preservation
- **Problem Solved**: SSO authentication was redirecting users to "/" instead of preserving the original `/chat/{blueprint_id}` URL
- **Solution**: 
  - Replaced Flask session-based OAuth state storage with in-memory, thread-safe dictionary (`_oauth_state_cache`)
  - Stores `(state, redirect_path)` mappings to preserve redirect URLs across SSO flow
  - Frontend passes current pathname as `redirect` query parameter
  - Backend retrieves and uses preserved redirect path after authentication

### 2. Auth Manager Changes (`auth_manager.py`)
- **State Management**: 
  - In-memory cache for OAuth state and redirect paths
  - Thread-safe operations for concurrent requests
  - Automatic cleanup of old states
- **Session Cookie Configuration**: 
  - Conditional `SESSION_COOKIE_SECURE` for development/production
  - `SESSION_COOKIE_SAMESITE` set to 'Lax' for cross-site redirects
  - Consistent secret key usage across server restarts

### 3. Frontend Auth Integration
- **Axios Interceptors**: Updated to capture and preserve redirect paths on 401 errors
- **AuthContext**: Includes current pathname in login redirect parameter

### 1. Invalid Workflow Validation
- **Before Sharing**: 
  - Attempts to create a test session (same as "Load Workflow")
  - Validates blueprint can be loaded, resolved, plan built, and graph compiled
  - Returns same error format as "Load Workflow" if validation fails
  - Prevents sharing toggle from being enabled on invalid workflows
- **Error Messages**: Consistent error messaging between "Load Workflow" and "Enable Sharing"

### 2. Disabled Sharing
- **On Active Chat**: 
  - Displays warning message: "Workflow Unavailable: This workflow's chat sharing has been disabled and can no longer be continued."
  - For maximum security, the sharing enablement is checked not only on rendering but when sending a new messages (for example, if a user has a chat page open, and the sharer of the blueprint disables it, we don't want the user to be able to continue chatting until he refreshes the page).
  - Disables chat input to prevent further interaction
  - No access to the public chat page anymore, if the user wants to see past conversations they can see them on the Agentic AI Chats page in UnifAI.
 Note that this applies to both shared link chats and normal agentic AI chats, and also for shared workflows that has been deleted.

### 4. Chat History Handling
- **User-Specific**: Only shows chat history for the current authenticated user and specific blueprint

### 1. Authentication Required
- All shared chat links require authentication via SSO
- User-specific chat history isolation

### 2. Ownership Validation
- Only blueprint owners can enable/disable sharing

### 3. Workflow Validation
- Invalid workflows cannot be shared

### DIFF
--- a/ui/client/src/components/agentic-ai/chat/PublicChat.tsx
+++ b/ui/client/src/components/agentic-ai/chat/PublicChat.tsx
@@ -6,6 +6,7 @@ import ChatInterface from "@/components/agentic-ai/chat/ChatInterface";
 import { SessionPayload } from "@/components/agentic-ai/ExecutionTab";
 import { StreamingDataProvider } from "@/components/agentic-ai/StreamingDataContext";
 import { Loader2, MessageSquare, Clock, Plus, Trash2, LogOut } from "lucide-react";
+import WorkflowStatusBanner, { WorkflowBannerMessages } from "@/components/shared/WorkflowStatusBanner";
 import { motion } from "framer-motion";
 import SimpleTooltip from "@/components/shared/SimpleTooltip";
 import { useTheme } from "@/contexts/ThemeContext";
@@ -75,12 +76,15 @@ export default function PublicChat() {
   }, []);
 
   // Check blueprint validity
-  const checkBlueprintValidity = useCallback(async (blueprintId: string) => {
-    setIsValidatingBlueprint(true);
+  const checkBlueprintValidity = useCallback(async (blueprintId: string, showLoadingState: boolean = true) => {
+    if (showLoadingState) {
+      setIsValidatingBlueprint(true);
+    }
     try {
       const result = await validateBlueprint({ blueprintId });
       setIsBlueprintValid(result.is_valid);
-      if (!result.is_valid) {
+      if (!result.is_valid && showLoadingState) {
+        // Only set validation error on initial load, not during polling
         setValidationError("Sorry, this workflow has validation errors and cannot be used. Please contact the workflow owner.");
       }
     } catch (error: any) {
@@ -88,7 +92,9 @@ export default function PublicChat() {
       // If validation fails, allow the chat to proceed but mark as potentially invalid
       setIsBlueprintValid(true); // Don't block on validation errors
     } finally {
-      setIsValidatingBlueprint(false);
+      if (showLoadingState) {
+        setIsValidatingBlueprint(false);
+      }
     }
   }, []);
 
@@ -136,19 +142,22 @@ export default function PublicChat() {
   // Load chat sessions when authenticated and blueprint is available
   useEffect(() => {
     if (isAuthenticated && user && blueprintId) {
-      // Check sharing status initially and periodically (every 30 seconds)
-      // This polling handles the scenario where the workflow owner DISABLES sharing while
-      // another user has the chat open. Without this, the user would continue chatting
-      // but their messages would fail with confusing errors. With polling, they'll see
-      // a clear "sharing disabled" message within 30 seconds of the owner disabling it.
+      // Check sharing status and blueprint validity initially and periodically (every 30 seconds)
+      // This polling handles scenarios where:
+      // 1. The workflow owner DISABLES sharing while another user has the chat open
+      // 2. The workflow becomes invalid (e.g., missing credentials, broken connections)
+      // Without polling, users would continue chatting but get confusing errors.
+      // With polling, they'll see clear messages within 30 seconds of changes.
       checkSharingStatus(blueprintId);
+      checkBlueprintValidity(blueprintId, false);
       const interval = setInterval(() => {
         checkSharingStatus(blueprintId);
+        checkBlueprintValidity(blueprintId, false);
       }, 30000);
       
       return () => clearInterval(interval);
     }
-  }, [isAuthenticated, user, blueprintId, checkSharingStatus]);
+  }, [isAuthenticated, user, blueprintId, checkSharingStatus, checkBlueprintValidity]);
 
   // Show loading state
   if (authLoading || isValidating) {
@@ -332,6 +341,16 @@ export default function PublicChat() {
                 <div className="text-center">
                   <Loader2 className="h-8 w-8 animate-spin mx-auto mb-4 text-primary" />
                   <p className="text-gray-400">Loading chat sessions...</p>
+                </div>
+              </div>
+            ) : !isBlueprintValid ? (
+              <div className="flex items-center justify-center h-full bg-background-dark">
+                <div className="max-w-md">
+                  <WorkflowStatusBanner
+                    variant={WorkflowBannerMessages.validationFailed.variant}
+                    title={WorkflowBannerMessages.validationFailed.title}
+                    message={WorkflowBannerMessages.validationFailed.message}
+                  />
                 </div>
               </div>
             ) : (
--- a/ui/client/src/components/agentic-ai/ExecutionTab.tsx
+++ b/ui/client/src/components/agentic-ai/ExecutionTab.tsx
@@ -316,6 +316,10 @@ export default function ExecutionTab({
           setChatSessions(prev => prev.map(s => 
             s.id === session.id ? { ...s, isSharingDisabled: disabled } : s
           ));
+          
+          // Validate blueprint for shared link sessions to check if workflow is still valid
+          // This catches cases where credentials are revoked or workflow has validation errors
+          await validateSelectedBlueprint(session.blueprintId);
         } else {
           // For non-shared-link sessions, sharing status doesn't matter
           setIsSharingDisabled(false);


odaiodeh:
@MayaCrmi 
I have reviewed the architecture regarding the "Shared Chat" feature. The current implementation treats "Sharing" as a separate entity/service, but logically, this is simply a state change on the Blueprint itself.

Here are the required changes to align the design with the Blueprint service architecture:

#### 1\. API Placement and Structure

**Issue:** The endpoints are currently placed in a `Share` service (`/api/shares/`) and modify the blueprint remotely.
**Fix:** Move these endpoints to the **Blueprint Service**. We are modifying a field on the blueprint, not creating a new resource.

**Refactoring Endpoints:**
Instead of four separate endpoints, please consolidate logic into **two management endpoints** and **one validation endpoint** under the `/blueprints` base path.

  * **Remove:** `/public-chat.enable`, `/public-chat.disable`, `/public-chat.status.get`
  * **Implement:**
    1.  **Set Visibility:** `PUT /blueprints/visibility`
          * **Payload:** `{ "blueprint_id": "...", "public": true }` (or `false`)
          * **Action:** Toggles the boolean on the document. One API handles both enabling and disabling.
    2.  **Get Visibility:** `GET /blueprints/visibility?blueprint_id=...`
          * **Action:** Returns current status.
    3.  **Validate:** `GET /blueprints/validate?blueprint_id=...`
          * **Action:** Checks if the blueprint exists and is public.

#### 2\. Separation of Concerns (Link Generation)

**Issue:** The backend is currently returning a constructed "Share Link."
**Fix:** The backend should **not** know about the shareable link structure.

  * The link is simply client-side logic (`Domain + Blueprint ID`).
  * The backend's only responsibility is to persist the state (`public: true/false`).
  * **Action:** Remove link generation logic from the API response. The UI will construct the link based on the `blueprint_id` and the successful `200 OK` response.

#### 3\. Naming Semantics

**Issue:** The naming convention `/public-chat` is misleading.
**Fix:**

  * We are changing the visibility of the Blueprint, not strictly the "chat."
  * Please use **Public / Private** terminology.
  * The field on the Blueprint is correctly named `shared` or `is_public` (prefer `public`).

#### 4\. Session Creation & Metadata

**Issue:** A new top-level parameter `from_shared_link` was added to `create_session`.
**Fix:**

  * **Do not pollute the top-level arguments.** Use the existing `metadata` dictionary parameter which is designed exactly for this purpose.
  * **Rename the key:** `from_shared_link` is verbose and tied to a specific UI flow.
  * **Proposed Change:** Pass this inside the metadata dict.
    ```json
    {
      "metadata": {
        "public": true
        // or "source": "public_link" here you can add link or what ever, since its UI injecting Metadata
      }
    }
    ```

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -64,36 +78,84 @@ def init_app(self, app):
             'PERMANENT_SESSION_LIFETIME': timedelta(hours=10)  # 10 hour sessions to match OIDC
         })
     
+    def _store_redirect_path(self, state: str, redirect_path: str):
+        """Store redirect path for a given OAuth state"""
+        with self._cache_lock:
+            self._oauth_state_cache[state] = (redirect_path, datetime.now())
+            logger.info(f"[STATE_CACHE] Stored redirect_path '{redirect_path}' for state '{state[:20]}...'")
```
Nirsisr: SSO_BACKEND is sharable util, which means like all UnifAI Users when they are interacting with the GUI, will reach over into the 'auth_manager.py' code, why do we need to print this one? 

You print here '{redireclt_path}' but how do we know for which user? It will appear under here 'state[:20]'? Like you are printing here any unique used_id or something indicating for which user we are storing this PATH?

Anyway still, we don't need those prints at least for now (if we had separate thread/logger per user it's something else).

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -64,36 +78,84 @@ def init_app(self, app):
...
+            # Extract state from session (authlib stores it there)
+            oauth_states = [key for key in session.keys() if key.startswith('_state_keycloak_')]
+            if oauth_states:
+                # Get the most recent state (the one just created)
+                latest_state_key = oauth_states[-1]
+                state = latest_state_key.replace('_state_keycloak_', '')
+                self._store_redirect_path(state, redirect_path)
+            
+            # Clean up expired states
+            self._cleanup_expired_states()
```
Nirsisr: I guess you already explained it on the general commit in the CR, but I will request for 1 0N 1 explanation on that part please. Just want to realize the logic and to make sure we are doing the right thing.

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -21,16 +19,32 @@ def __init__(self, app=None):
...
+                # In production, secret_key is required for security
+                if self.backend_env == 'production':
+                    raise ValueError(
+                        "secret_key must be configured in production environment. "
+                        "Set 'secret_key' in your configuration to prevent security vulnerabilities."
+                    )
+                # Development fallback (only allowed in non-production)
+                secret_key = 'dev-secret-key-fixed-for-session-persistence-change-in-production'
+                logger.warning("Using default development secret key - set 'secret_key' in config for production!")
+            app.secret_key = secret_key
```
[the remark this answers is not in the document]
MayaCrmi: Removed all environment-based conditional logic that I added. I just extract the secret_key now and kept the fallback to a default value if we can't find it.

shared-resources/sso-backend/utils/auth_manager.py
MayaCrmi: Removed logging statements that don't identify which user the log is for. Kept only logs that include user identification.

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -64,36 +78,84 @@ def init_app(self, app):
...
+            if state in self._oauth_state_cache:
+                redirect_path, timestamp = self._oauth_state_cache[state]
+                # Check if entry is expired (more than 15 minutes)
+                if datetime.now() - timestamp < timedelta(minutes=15):
+                    if remove_after_use:
+                        del self._oauth_state_cache[state]
+                    return redirect_path
+                else:
+                    del self._oauth_state_cache[state]
+            return '/'
```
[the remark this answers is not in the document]
MayaCrmi: The method retrieves the user's intended destination path that was stored during login initiation (what we received from the original URL auth_callback() retrieves the redirect path after OAuth callback to redirect user to their intended destination.

The 15-minute keeping is what was suggested, though in practice the OAuth flow takes like only a few seconds (I can change it to less time). It's recommended since it provides a safety net for when the network is slow or if the user doesn't immediately log in (like when you come to log in in SSO and it says it's not valid anymore and you need to refresh to log in?). Anyway it's automatically cleaned up after use or expiration.

The flow here is the user logins from 'some-url', then _store_redirect_path() stores it mapped to OAuth state so that after authentication Keycloak redirects back with the same OAuth state.  _get_redirect_path() retrieves 'some-url' using that state and redirects to UI with this path.

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -64,36 +78,84 @@ def init_app(self, app):
...
-            # redirect_uri = config.get(
-            #     'redirect_url',
-            #     url_for('auth_callback', _external=True, _scheme='https')
-            # )
+            # Get the redirect URL from query parameter (where to send user after auth)
+            redirect_path = request.args.get('redirect', '/')
+            # Ensure redirect_path starts with / and is safe
+            if not redirect_path.startswith('/'):
+                redirect_path = '/'
+            logger.info(f"[LOGIN] Redirect path requested: {redirect_path}")
```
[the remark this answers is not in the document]
MayaCrmi: Removed this.

shared-resources/sso-backend/utils/auth_manager.py
MayaCrmi: Sure, we'll discuss F2F but here's a short explanation:
authorize_redirect() is called, authlib generates a random OAuth state string for CSRF protection (to prevent that CSRF error we sometimes saw) so authlib stores this state in the Flask session under a key. Then, we extract all session keys starting with the relevant prefix, go to the one we just created, and then extract the state value by removing said prefix. Then last we store a mapping like 'state -> redirect_path' in our cache.

We need this since OAuth state is required for security (CSRF) and we cannot modify the state to include redirect path (as said, it's a security risk). So to work around this, we need to map the state to redirect_path so we can retrieve it after callback. In general, authlib manages the state, we just extract it to create our mapping

In the beginning I tried passing redirect_path as a custom parameter, but this is less secure (and also from reading now it also seems that it's not standard OAuth practice).

I asked the cursor to add here more details in the documentation of this part, hopefully this is helpful.

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -114,20 +176,28 @@ def auth_callback():
...
-                frontend_url = config.get('frontend_url', 'http://localhost:5000')
-                return redirect(f"{config.frontend_url}?auth=error")
-
+                
+                request_state = request.args.get('state', '')
+                redirect_path = self._get_redirect_path(request_state, remove_after_use=False) if request_state else '/'
+                
+                redirect_url = f"{config.frontend_url}{redirect_path}?auth=error"
+                if redirect_path != '/':
+                    redirect_url += f"&redirect={redirect_path.replace('/', '%2F')}"
```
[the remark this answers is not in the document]
MayaCrmi: NP, I  added a function _get_redirect_path_encoded() that uses `urllib.parse.quote()` for proper URL encoding. This will handle all URL encoding properly, not just /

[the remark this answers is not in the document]
Nirsisr: Cool, will go over it.

multi-agent/api/flask/endpoints/blueprints.py
```diff
@@ -34,9 +34,10 @@ def available_resolved_doc_list(user_id):
 @blueprints_bp.route("/blueprint.save", methods=["POST"])
 @from_body({
     "blueprint_raw": fields.Str(data_key="blueprintRaw", required=False),  # optional for non-JSON/YAML raw
-    "user_id": fields.Str(data_key="userId", required=False, load_default="alice")
+    "user_id": fields.Str(data_key="userId", required=False, load_default="alice"),
+    "metadata": fields.Dict(data_key="metadata", required=False, load_default=lambda: {}, dump_default=lambda: {})
```
odaiodeh: what is this dump_default and what does it do?

multi-agent/api/flask/endpoints/blueprints.py
```diff
@@ -134,3 +155,77 @@ def remove_blueprint(blueprint_id):
...
+            "error": str(e),
+            "error_type": "INVALID_METADATA",
+            "blueprint_id": blueprint_id
+        }), 400
+    except Exception as e:
+        return jsonify({
+            "error": str(e),
+            "error_type": "VALIDATION_ERROR",
+            "blueprint_id": blueprint_id
+        }), 400
```
odaiodeh: i dont think these exception are really meaningful, like KeyError doesnt mean BLUEPRINT_NOT_FOUND, and ValueError doesnt mean INVALID_METADATA and the same for generic Exception, exceptions should be generic, look at other endpoints in other services how its done.

multi-agent/api/flask/endpoints/sessions.py
```diff
@@ -19,7 +19,7 @@ def create_user_session(blueprint_id, user_id, metadata):
         session_svc = current_app.container.session_service
         session = session_svc.create(user_id=user_id,
                                      blueprint_id=blueprint_id,
-                                     metadata=metadata)
+                                     metadata=metadata or {})
```
odaiodeh: you already did load_default=lambda: {}, why you need to do or {}

multi-agent/blueprints/repository/mongo_blueprint_repository.py
odaiodeh: removing the whole file and then adding it again is not good, since i don't see the specific change you did, and in the future for git blame its not good.
please re push it with the specific change, so i see in diff the only change.

multi-agent/sharing/service.py
```diff
@@ -14,9 +15,11 @@ class ShareService:
     
     def __init__(self, 
                  share_repository: ShareRepository,
-                 cloner: ShareCloner):
+                 cloner: ShareCloner,
+                 session_service: SessionService = None):
         self._repo = share_repository
         self._cloner = cloner
+        self._session_service = session_service
```
odaiodeh: why is it needed here now?

multi-agent/session/service.py
```diff
@@ -122,4 +134,28 @@ def delete(self, run_id: str) -> bool:
...
+        except Exception as e:
+            raise Exception(f"Blueprint validation failed: {str(e)}") from e
+        finally:
+            # Clean up test session if it was created
+            if test_session:
+                try:
+                    self.delete(test_session.get_run_id())
+                except Exception:
+                    # Ignore cleanup errors
+                    pass
```
odaiodeh: i don't like this approach of validating blueprint by making test session then deleting it, we don't still have the write validating method, im working now on some epic of validation, that we can use it after i do it, but this way of validate_blueprint inside a session service? not good.

multi-agent/session/dto.py
```diff
@@ -9,16 +9,18 @@ class ChatHistoryItem:
...
+    def from_doc(cls, doc: Mapping[str, Any], blueprint_exists: bool = True, public_usage_scope: bool = False) -> "ChatHistoryItem":
         rc = doc.get("run_context", {})
         return cls(
             session_id=rc.get("run_id"),
             metadata=doc.get("metadata", {}),
             started_at=rc.get("started_at"),
             blueprint_id=doc.get("blueprint_id", ""),
-            blueprint_exists=blueprint_exists
+            blueprint_exists=blueprint_exists,
+            public_usage_scope=public_usage_scope
```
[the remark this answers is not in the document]
MayaCrmi: gotcha, so keeping it inside the sessions to be needed but under metada ok, changing it

multi-agent/session/service.py
odaiodeh: lets finish this PR after the validation EPIC which will be done these these two days, and then we use the validation from there.

ui/client/src/api/blueprints.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Blueprint Metadata & Sharing
+// ────────────────────────────────────────────────────────────────────────────────
+
+/**
+ * Check if a blueprint has public sharing enabled
+ */
+export async function getPublicUsageScope(blueprintId: string): Promise<{ public_usage_scope: boolean; blueprint_id: string }> {
```
Nirsisr: @MayaCrmi Just wonder why we need separate API for this one, why we are not having the data of it under `getBlueprintInfo` I mean that could reduce the number of API calls, or am I missing something?

ui/client/src/api/blueprints.ts
Nirsisr: @MayaCrmi 10x for putting all the logic here, great work <3

ui/client/src/components/agentic-ai/ShareWorkflow.tsx
```diff
@@ -0,0 +1,216 @@
...
+      } catch (error) {
+        console.error("Error fetching public_usage_scope status:", error);
+      }
+    };
+
+    fetchStatus();
+  }, [blueprintId]);
+
+  const handleToggle = async (checked: boolean) => {
+    if (!user?.username) {
```
Nirsisr: This senario is even possible Maya? I mean if the user first using `{shared_session_url}` has is first landing page, he will ever land on this condition? I mean what I want to believe will happen is re-direction directly to SSO Auth page no matter on what's is the UnifAI URL you are putting.

ui/client/src/components/agentic-ai/chat/PublicChat.tsx
```diff
@@ -0,0 +1,394 @@
...
+        <div className="text-center">
+          <Loader2 className="h-8 w-8 animate-spin mx-auto mb-4 text-primary" />
+          <p className="text-gray-400">Loading...</p>
+        </div>
+      </div>
+    );
+  }
+
+  // Show authentication required
+  if (!isAuthenticated) {
```
Nirsisr: @MayaCrmi Same question as I asked above, wouldn't the user will be re-directed to SSO page immediately when he's landing on each UNIFAI URL? Even if it's internal URL? with `/shared/XXX` or whatever?

ui/client/src/components/agentic-ai/chat/PublicChat.tsx
```diff
@@ -0,0 +1,394 @@
...
+        </div>
+      </div>
+
+      {/* Delete Chat Confirmation Modal */}
+      <AlertDialog open={showDeleteModal} onOpenChange={setShowDeleteModal}>
+        <AlertDialogContent className="bg-background-card border-gray-800">
+          <AlertDialogHeader>
+            <AlertDialogTitle>Delete Chat</AlertDialogTitle>
+            <AlertDialogDescription>
+              Are you sure you want to delete "{chatToDelete?.title}"? This action cannot be undone.
```
Nirsisr: @MayaCrmi On the PublicChat also user can press like + btn to have multiple chats with the same workflow? Just wonder how it work. Since if it's 1 Chat, what's the delete btn means here? (if you are allowing duplicate the same workflows multiple times so it's another story of course)

ui/client/src/contexts/AuthContext.tsx
Nirsisr: @MayaCrmi after we are merging it, please invite 15 min the entire team for a discussion on the changes over here, those are so crucial and we have some back and forth till reaching out for this current solution.

**This one super important to just make everyone in the same picture and aware for those changes.**

ui/client/src/types/session.ts
```diff
@@ -0,0 +1,33 @@
+// Backend message format
+export interface BackendChatMessage {
```
Nirsisr: About the naming here, where mentioned in previous comment (just put it here so you won't forget to change)

ui/client/src/utils/index.ts
```diff
@@ -58,4 +58,23 @@ export function formatDate(dateStr: string): string {
     minute: '2-digit',
     hour12: false
   });
+};
+
+/**
+ * Format timestamp as relative time (e.g., "Just now", "5m ago", "2h ago", "3d ago")
+ * Falls back to date string for timestamps older than 7 days
+ */
+export function formatRelativeTimestamp(timestamp: string): string {
```
Nirsisr: Regard this one, I remember the exact same function (maybe it's on on-going PR, yosi for statistics, I remember the exact same one :D) Just want to make sure we don't have the same helpers across all the app

ui/client/src/api/agentic.ts
```diff
@@ -1,14 +1,13 @@
...
-import { BlueprintValidationResult, BlueprintValidationRequest } from '@/types/validation';
 
-export interface WorkflowBlueprint {
-  blueprint_id: string;
-  spec_dict: any;
-  name?: string;
-  created_at?: string;
-  updated_at?: string;
-}
+// Re-export blueprint functions for backwards compatibility
```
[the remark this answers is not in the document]
MayaCrmi: it's so that the existing imports like fetchBlueprints/WorkflowBlueprint from /api/agentic continue to work without breaking changes, since they were originally in agentic.ts but are now in the new blueprints file, becuase without this we'd need to update like 6-7 files that currently import from agentic.. we can change all of this if you'd prefer

ui/client/src/api/blueprints.ts
MayaCrmi: yeah you're right, it's more for convenience and clarity. what happens is that getPublicUsageScope actually calls getBlueprintInfo and gets the usageScope, so there's no extra api call. we could simplify by having it just call getBlueprintInfo directly if you prefer, but i think it's nicer to have like
getPublicUsageScope(id)
instead of
getBlueprintInfo(id)).metadata.usageScope

ui/client/src/components/agentic-ai/chat/PublicChat.tsx
```diff
@@ -0,0 +1,394 @@
...
+    validateToken();
+  }, [token, checkSharingStatus, checkBlueprintValidity]);
+
+
+  // Load chat sessions when authenticated and blueprint is available
+  useEffect(() => {
+    if (isAuthenticated && user && blueprintId) {
+      // Check sharing status initially and periodically (every 30 seconds)
+      checkSharingStatus(blueprintId);
+      const interval = setInterval(() => {
```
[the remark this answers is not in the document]
MayaCrmi: yes exactly, like every 30 seconds it's called to handle the scenario where the workflow owner disables sharing while someone has the chat open. so instead of the user starting to get failure messages with confusing errors, the user will see a clear message stating that the sharing was disabled. it's a pretty minimal api call every half minute while the chat is active so it seemed like the gain is greater than the loss IMO

ui/client/src/contexts/AuthContext.tsx
MayaCrmi: yeah absolutely no problem!

MayaCrmi: i can't comment directly on this message:

image

So responding here:
there's noo change in behavior - like unauthenticated users will still be redirected to SSO automatically. the only change is that after SSO login, they'll return to their original URL that sent them to the login instead of always landing on /. so in your example you'll land back on inventory.
the if isAuthenticate checks in PublicChat etc are just fallbacks but we can remove them as I said if they cause any confusions and unneeded noise in your opinion

ui/client/src/hooks/use-session-management.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Custom hook for session management operations
+ */
+
+import { useState, useCallback } from 'react';
+import axios from '@/http/axiosAgentConfig';
+import { ChatSession, SessionStateData, BackendChatMessage } from '@/types/session';
```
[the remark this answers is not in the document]
MayaCrmi: NP, changing them all, so renaming BackendChatMessage to ChatMessage across all the files I added in the helpers, types etc

ui/client/src/hooks/use-sharing-status.ts
```diff
@@ -0,0 +1,77 @@
...
+    checkSharingStatus,
+    resetStatus,
+  };
+};
+
+/**
+ * Check sharing status for a session (with fallback)
+ * Returns the sharing disabled status, using API response as fallback
+ */
+export const checkSessionSharingStatus = async (
```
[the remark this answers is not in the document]
MayaCrmi: ok so it's in two cases called: the first is the hook uneder this files, so that's used when the component needs to re-render based on status changes and the second is the direct function that we discussed earlier like getscope that's  only used for one-time checks in where you just need the value as it and not like be reactive

MayaCrmi:
@Nirsisr alll of your comments have been answered, let me know your thoughts on the other side (:

sfiresht:
I see that there are changes which are written in the MR description but are not relevant anymore. Not sure you need to change it though :-)

MayaCrmi:
> I see that there are changes which are written in the MR description but are not relevant anymore. Not sure you need to change it though :-)

this is like the 6th cycle of the CR  hhh

ui/client/src/components/agentic-ai/chat/ChatInterface.tsx
```diff
@@ -685,28 +711,32 @@ export default function ChatInterface({
...
-              className="text-gray-400 hover:text-gray-100"
-              title={isBlueprintGraphHidden ? "Show Blueprint Graph" : "Hide Blueprint Graph"}
-            >
-              {isBlueprintGraphHidden ? (
-                <ChevronLeft className="h-4 w-4" />
-              ) : (
-                <ChevronRight className="h-4 w-4" />
+          {!isChatOnlyMode && (
+            <>
+              <Button
```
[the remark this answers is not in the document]
MayaCrmi: added these four:
WORKFLOW_SHARE_ENABLE = "workflow-share-enable",
    WORKFLOW_SHARE_DISABLE = "workflow-share-disable",
    WORKFLOW_SHARE_COPY_LINK = "workflow-share-copy-link",
    PUBLIC_CHAT_NEW_SESSION = "public-chat-new-session",
if you want to take a look and see if looks good.

multi-agent/session/workflow_session_factory.py
```diff
@@ -44,11 +44,15 @@ def create(
...
+        # Ensure metadata is a SessionMeta instance
+        session_meta = metadata if metadata is not None else SessionMeta()
+        
         # 0) Build and propagate RunContext ———
+        # RunContext.metadata expects a dict, so convert SessionMeta to dict
         ctx = RunContext(
             user_id=user_id,
             engine_name=self._engine_name,
-            metadata=metadata or {}
+            metadata=session_meta.to_dict()
```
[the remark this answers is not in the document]
sfiresht: so why do we need  .to_dict here?

ui/client/src/contexts/AuthContext.tsx
```diff
@@ -76,15 +84,51 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
     const urlParams = new URLSearchParams(window.location.search);
     const authStatus = urlParams.get('auth');
+    const stateParam = urlParams.get('state');
     
     if (authStatus === 'success') {
-      // Remove auth params from URL
-      window.history.replaceState({}, document.title, window.location.pathname);
+      // Extract original URL from state parameter
+      let originalUrl = '/';
+      if (stateParam) {
```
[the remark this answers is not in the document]
Nirsisr: Maya this one is happen after user save his blueprint (that he just built the graph for)? Or AM I missing between stuff? Like maybe explain to me in short 1 0N 1 after we closed all the stuff.

So not sure I got why `stateParam is missing for any reason` like I felt this case can't never happen (with the current code imp) or maybe I don't understand the flow well.

ui/client/src/contexts/AuthContext.tsx
Nirsisr: It's 100% unneeded noise, which I don't see any reason to have this code who SHOULD NEVER be called.

10x for the explanation, just raise it in the team meeting. Great. (you can resolve when you close the fallbacks)

ui/client/src/hooks/use-public-chat.ts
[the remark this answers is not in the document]
Nirsisr: I'm ok with that, please open a ticket with the entire tickets we just discussed here and we can tackle it later.
You can resolve once done, 10x.

ui/client/src/components/agentic-ai/ShareWorkflow.tsx
MayaCrmi: ok, removed the if check

ui/client/src/contexts/AuthContext.tsx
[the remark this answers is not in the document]
MayaCrmi: like stateParam saves the original URL through the SSO flow, and the fallback originalUrl = '/' handles edge cases like user bookmarked the SSO callback URL directly like manual URL manipulation. in the code in general stateParam will always be present and the fallback is defensive - if somehow state is missing or something or user goes to landign page instead of getting an error

ui/client/src/api/blueprints.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export interface WorkflowBlueprint {
+  blueprint_id: string;
+  spec_dict: any;
+  name?: string;
+  created_at?: string;
+  updated_at?: string;
+}
+
+export interface BlueprintInfoResponse {
```
Lina-AbuYousef: And about the “response interfaces”: I actually think this is the right place for them, since they’re specific to the expected response shape (API contract) for this module/endpoint.

MayaCrmi: agree (:

ui/client/src/api/agentic.ts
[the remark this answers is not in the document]
Nirsisr: 10x, resolving.

ui/client/src/hooks/use-public-chat.ts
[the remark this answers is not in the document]
Nirsisr: For sure you can resolve this one, once you open the tickets. 10x

Nirsisr:
> @Nirsisr commented on second cycle, please resolve any comments that need resolving and continue the discussion where you find necessary. here's a list of follow-ups I will do in continuation of this branch once you approve I understand them correctly etc:
> 
> 1. Session Management Hooks Consolidation (open a ticket):
>    The use-public-chat hook contains reusable session management logic that could be used in ExecutionTab as well. Currently, ExecutionTab has its own inline implementation of similar functionality. We want to create a generic use-session-chat hook that can be used by ExecutionTab and PbulicChat in relation to each of their funcionalities.
> 2. Review time formatting functions (for the UI Maintainers Session):
>    During code review, Nir raised the concern that we have multiple time formatting functions. so we want to add this discussion to GUI maintainers session agenda to review together and reduce duplication.
> 3. Helpers folder organization:
>    Not quite sure if this should be a ticket or a discussion or a discussion and then a ticket. We want to discuss whether component-specific helpers (like sessionHelpers should live in a utils folders like no, or like components/component/helpers or inside hooks if they're used there. What do you think?

@MayaCrmi Perfect job with this summarization. Follow up to this I resolved all of my comments. 10x

ui/client/src/api/blueprints.ts
Nirsisr: Totally agree with that.

Nirsisr (dismissed):

ui/client/src/hooks/use-public-chat.ts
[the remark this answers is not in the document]
MayaCrmi: opened and resolved

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -28,9 +26,12 @@ def init_app(self, app):
         """Initialize the auth manager with Flask app"""
         self.app = app
         
-        # Set up secret key for sessions
+        # Set up secret key for sessions (required for secure session management)
+        # The secret_key should be configured in app_config for both dev and production
+        # A permanent key ensures sessions persist across container restarts
         if not app.secret_key:
-            app.secret_key = config.get('secret_key', os.urandom(24))
+            app.secret_key = config.get('secret_key')
```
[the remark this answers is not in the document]
MayaCrmi: kept the random part here and we can discuss more in future if needed?

odaiodeh (approved):

Nirsisr (dismissed):

odaiodeh (dismissed):

Nirsisr (approved):

PR #8, opened 2025-11-25
Genie 800/bug/display names instead of UUID

MayaCrmi:
# Display Resource Names Instead of UUIDs in Agents Screen

Under the "Agents" screen (User Workspace), when viewing agent configurations, the system was displaying UUIDs instead of human-readable names for referenced resources (LLMs, providers, tools). This made it difficult for users to understand which resources their agents were using.

When discussing this ticket with Nir he instructed me to start with creating a centralized `AgenticAIContext` that maintains, among other things, a mapping between resource UUIDs and their names across all user inventory elements. 

### New Features
- **AgenticAIContext** (`ui/client/src/contexts/AgenticAIContext.tsx`)
  - Fetches all user inventory resources from the backend
  - Maintains UUID → name and UUID → resource mappings
  - Provides utility functions:
    - `getResourceName(ref)` - Resolves a single ref to its name
    - `getResource(ref)` - Gets full resource info for a ref
    - `resolveRefsInConfig(config)` - Recursively resolves all refs in a config object

Using these functions, components like ElementGrid and ElementData now display resource names instead of UUIDs, as expected.

### DIFF
--- a/ui/client/src/App.tsx
+++ b/ui/client/src/App.tsx
@@ -45,7 +45,7 @@ function App() {
                   <Route path="/" component={withAgenticAIProvider(AgenticAI)} />
                   <Route path="/agentic-ai" component={withAgenticAIProvider(AgenticAI)} />
                   <Route path="/inventory" component={withAgenticAIProvider(AgentRepository)} />
-                  <Route path="/agentic-chats" component={AgenticChats} />
+                  <Route path="/agentic-chats" component={withAgenticAIProvider(AgenticChats)} />
                   
                   {/* Non-agentic routes - don't need AgenticAIProvider */}
                   <Route path="/jira" component={JiraIntegration} />


ui/client/src/App.tsx
```diff
@@ -27,27 +28,29 @@ function App() {
...
+                    <Route path="/slack/add-source" component={SlackAddSourcePage} />
+                    <Route path="/configuration" component={Configuration} />
+                    <Route path="/guides" component={GuidesPage} />
+                    <Route component={NotFound} />
+                  </Switch>
+                </ProtectedRoute>
+              </NotificationProvider>
+            </AgenticAIProvider>
+          </ProjectProvider>
+        </SharedProvider>
```
Nirsisr: We already discussed this one F2F, we are both in the same picture here unfortunately. 

AMMM as it seems we are going to return the RAG part at least for docs, so please try to wrap only the agentic-ai parts with the AgenticAIProvider, maybe split the route to few groups and wrap only the relevant wrap with the relevant provider, try to understand how it's being DONE, not sure about it.

SharedPrvodier, NotificationsProvider are both relevant for all the app (tool bar attached) so those are fine to wrap all of them.

MayaCrmi: The <Switch> wrapper didn't support me putting the context straight in so added a component to call with the wrapper, only taking the agentic AI components into account.

Nirsisr: That's a good workaround, solved.

ui/client/src/contexts/AgenticAIContext.tsx
```diff
@@ -0,0 +1,228 @@
...
+      setUuidToNameMap(nameMap);
+      setUuidToResourceMap(resourceMap);
+    } catch (err: any) {
+      const errorMessage = err.response?.data?.error || "Failed to fetch resource mappings";
+      setError(errorMessage);
+      console.error("Error fetching resource mappings:", err);
+    } finally {
+      setIsLoading(false);
+    }
+  }, [USER_ID]);
```
Nirsisr: @MayaCrmi I think you are missing here something, ammmm what if user is adding new element in his catalog? We need to update our 'UID:NAME' Mapping on the fly, immediately. 

Imagine user create an LLM, than edit an agent (selecting the newly created LLM for that agent), if you are not updating the lists, the LLM new ref won't be there.

So each addition/deletion of an element under User Inventory should trigger an update to fetchAllResources in a smart way, I won't accept for a call to fetch all the resources from scratch but something smarter. (it just came to my mind know, and I guess that it will raise a bug on your implementation)

ui/client/src/hooks/use-workspace-data.ts
```diff
@@ -350,7 +373,7 @@ export const useWorkspaceData = () => {
         setIsLoading(false);
       }
     },
-    [toast],
+    [toast, addOrUpdateResource, USER_ID],
```
Nirsisr: Please see my comment from below regards the deps array, 'addOrUpdateResource' should never changed since it's one time created function therefore I don't see any reason to put it here in the deps.

Nirsisr (dismissed):

sfiresht (commented):
did you check it yet on preprod?

PR #10, opened 2025-11-27
Genie 1001/story/one time chat option

Lina-AbuYousef:
Pls don't CR this one yet, I just open an PR for myself its 90%. ready but still need small stuff.
Will reopen it again once I finish with more important ticket before.


Nirsisr:
@Lina-AbuYousef Please add description to the CR before we are going over it Linas, what the purpose of the CR? What you are trying to achieve here? Relevant attachment for JIRA ticket, etc. 

I have no idea about this ticket at all TBH, so need to have some information before digging in.

PR #11, opened 2025-11-30
Genie 861 new umami

sfiresht:
This PR introduces integration with Umami UI monitoring system into UnifAI.
In this PR:

- Add settings endpoint in backend to fetch Umami configuration (and future settings needs)
- Inject Umami script dynamically in frontend index.html
- Add event tracking attributes to primary UI actions + session tracking
- Update Helm charts with Umami environment variables
- small helm files fixes (Adjust PVC protection location ,probe timings indentation)

How does the Umami flow work:
1. The user browses to unifai and gets the UI client files
2. the "index.html" file includes a call to an external script in Umami with the route and specific website ID
3.the user browses and the scripts invokes sends of the current user pages in addition we've added specific triggers to get more information about the users' behavior.
4. the umami gets the data and saves it to its DB.

the Umami UI resides at: http://umami-route-tag-ai--sfiresht-nb.apps.stc-ai-e1-pp.imap.p1.openshiftapps.com/dashboard

**NOTE: I know that the triggers at the moment may not be the ones everyone agrees on so please comment and suggest things that we need to have as triggers.**

### DIFF
--- a/ui/client/src/components/agentic-ai/graphs/SaveBlueprintModal.tsx
+++ b/ui/client/src/components/agentic-ai/graphs/SaveBlueprintModal.tsx
@@ -1,5 +1,5 @@
 
-import React, { useState } from "react";
+import React, { useState , useMemo} from "react";
 import {
   Dialog,
   DialogContent,
@@ -44,6 +44,7 @@ const SaveBlueprintModal: React.FC<SaveBlueprintModalProps> = ({
 
   const isFormValid = name.trim().length > 0;
   const { user } = useAuth();
+  const timestamp = useMemo(() => new Date().toISOString(), [isOpen]);
 
   return (
     <Dialog open={isOpen} onOpenChange={handleClose}>
@@ -99,6 +100,7 @@ const SaveBlueprintModal: React.FC<SaveBlueprintModalProps> = ({
             data-umami-event="agent-graphs-save-blueprint-button" 
             data-umami-event-user-id={user?.username}
             data-umami-event-blueprint-name={name}
+            data-umami-event-blueprint-unique-id={`${name}-${timestamp}`}
           >
             {isLoading ? (
               <div className="flex items-center gap-2">
--- a/ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
+++ b/ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useMemo} from "react";
 import {
   Dialog,
   DialogContent,
@@ -65,6 +65,7 @@ export const ElementForm: React.FC<ElementFormProps> = ({
   };
 
   const { user } = useAuth();
+  const timestamp = useMemo(() => new Date().toISOString(), [isOpen]);
 
   // Initialize form data
   useEffect(() => {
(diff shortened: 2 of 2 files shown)


backend/endpoints/settings.py
```diff
@@ -0,0 +1,16 @@
+from flask import Blueprint, jsonify
+from shared.logger import logger
+from providers.settings import get_umami_settings as _get_umami_settings
+
+# we might want this as a "settings bp and not umami only"
```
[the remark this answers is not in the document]
Nirsisr: I'm favor with what Saar said, like general endpoint for any integrations we have for UnifAI, Umami is just one in the list if we have more in the future we can add them here.

ui/client/src/components/agentic-ai/chat/ChatInterface.tsx
```diff
@@ -756,6 +759,9 @@ export default function ChatInterface({
               onClick={handleSendMessage}
               disabled={inputMessage.trim() === "" || isTyping || !blueprintExists}
               className="bg-primary hover:bg-[#7525c9] mb-0"
+              data-umami-event="agent-chat-send-message-button" 
+              data-umami-event-user-id={user?.username}
+              data-umami-event-flow-name={runId}
```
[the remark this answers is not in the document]
Nirsisr: It's stand for the sessionID maya

ui/client/src/components/agentic-ai/graphs/SaveBlueprintModal.tsx
```diff
@@ -94,6 +96,9 @@ const SaveBlueprintModal: React.FC<SaveBlueprintModalProps> = ({
             onClick={handleSave}
             disabled={!isFormValid || isLoading}
             className="bg-primary hover:bg-primary/80"
+            data-umami-event="agent-graphs-save-blueprint-button" 
+            data-umami-event-user-id={user?.username}
+            data-umami-event-blueprint-name={name}
```
Nirsisr: In this case we will have a table for each username with all his blueprints names, and number of counts per each? I would send 'blueprint-id' also since it's 100% unique identifier, like it rare but can happen that 2 blueprints have exactly the same name (so it problematic to base the tracking only on that)

ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
```diff
@@ -672,6 +673,9 @@ export const ElementForm: React.FC<ElementFormProps> = ({
               type="submit"
               className="bg-primary hover:bg-opacity-80"
               disabled={isSaving || !isFormValid()}
+              data-umami-event="agent-repository-save-element-button" 
+              data-umami-event-user-id={user?.username}
+              data-umami-event-element-type={elementType.name}
```
Nirsisr: Same comment as above, different elements can have the same name, so we need to track the UID (I wrote those comments also in the prev CR in gitlab, this one is tricky)

ui/client/src/components/agentic-ai/graphs/SaveBlueprintModal.tsx
sfiresht: good thinking, but since we can't map id to name it will not be so readable, maybe if we add both id and name it will better?, I'll add the id in addition to what we have now :-)

sfiresht: created a new property - unique-id combining name-timestamp.
update me if it's ok or somthing else needed.

sfiresht:
> Just a general question, now when adding features (for example when I add now the Chat Only option and a user can click "allow sharing" or something) will I also need to add the umami to my buttons etc? Do we need to make sure everyone's aware of their responsibility of tracking any needed data etc?

well, it depends on the feature. we don't want to track each button click, right? in the chat only, there's a chance that we still want to track it as it includes information on which flow the user chose.

PR #12, opened 2025-11-30
move  tokens from app_config to git_creds repo

yhabushi79:
move slack credentials to k8s secrets

- Remove hardcoded Slack tokens from backend config
- Add helm presync hook to generate unifai-dataflow-secrets
- Inject secrets via envFrom in Server and Celery deployments
- Improve configmap deletion idempotency

These values are now injected into the system through the Helmfile pod as environment variables. During deployment, the Helmfile pod passes these environment variables into the Dataflow pods, ensuring the tokens remain secure and are never stored in the public repo ( under app_config.py)


odaiodeh (approved):

helm/dataflow-presync.sh
```diff
@@ -0,0 +1,86 @@
...
+    log_warn "Make sure to set these variables if Slack integration is required."
+fi
+
+# Create Secret with Slack tokens from environment variables
+log_info "Creating/updating Secret 'unifai-dataflow-secrets'..."
+
+SECRET_NAME="unifai-dataflow-secrets"
+
+# Check if secret already exists
+if kubectl get secret "$SECRET_NAME" &>/dev/null; then
```
sfiresht: the &>/dev/null here for a case where the secret isn't found?
if so I think we can use this --ignore-not-found=true (as you already used below)

PR #13, opened 2025-11-30
removed unnecessary files + organized existing onesin scripts

sfiresht:
organizing helmfile hooks to reusable scripts

- Move repeated wait logic to `helm/scripts/common.sh`
- Extract service-specific postsync hooks to dedicated scripts
- Refactor .gotmpl files to execute script files instead of inline bash
- Clean up obsolete `helmfile*.yaml` and `postsync.sh` files
- Improve error handling: hooks now fail fast on timeout instead of applying invalid configs

### DIFF
--- a/helm/scripts/common.sh
+++ b/helm/scripts/common.sh
@@ -72,7 +72,7 @@ wait_for_ip() {
 create_or_update_configmap() {
   local cm_name=$1
   shift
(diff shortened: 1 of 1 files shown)


sfiresht:
merged with main and tested on pre-prod

PR #21, opened 2025-12-04
Document UI, management, rendering improvements

MayaCrmi:
# PR Description: Bulk Delete and Document Management Improvements

## 🎯 Overview
This PR includes bulk delete functionality for Documents, along with significant improvements to document upload validation, file handling, and overall UX. The implementation follows a modular architecture with reusable hooks and components.

*Note that I haven't done it for Slack as well since it's still irrelevant. If needed we can add it in the future.*

### Bulk Delete Functionality
Added checkbox selection in both grid and list views for documents using `useBulkDelete` hook for handling batch operations. Added a new `BulkDeleteButton` component with selection count display.

image

### Document Upload Improvements
Enforces 50MB limit with clear error messages with dynamic validation based on backend-supported file extensions. Improved error messages for invalid files and the list update after successful upload (now happens immediately). As requested, also added a check that the user doesn't have a file by the same name in his db yet to not allow duplicates.

image

### Backend Improvements
- **Secure Filename Handling**: Uses `secure_filename` to properly handle filenames with spaces and special characters
- **.pptx Support**: Added support for PowerPoint files (.pptx)
- **Consistent File Handling**: Registration logic matches upload logic for filename handling
- Removal of the uploaded file after the pipeline is done.

## 🐛 Bugs Fixed
- Fixed issue with documents having spaces in title (handled with `secure_filename`)
- Fixed type safety issues in DataCard component
- Fixed delete mechanism so that the confirmation modal stays present until the deletion finishes.

## Misc
- I have already added a small addition to the google mcp script that we were requested to add in our channel.
- We're using the SlackIntegration file from under /features but we still had it also under /pages so I deleted it to prevent confusion.

### DIFF
--- a/scripts/cleanup_slack_retrievers.py
+++ b/scripts/cleanup_slack_retrievers.py
@@ -1,0 +1,342 @@
+#!/usr/bin/env python3
+"""
+Find and clean up Slack retrievers and their associated blueprints.
+
+Usage:
+    # Dry run - list all slack retrievers and blueprints using them
+    python cleanup_slack_retrievers.py
+    
+    # Apply changes - delete slack retrievers and blueprints
+    python cleanup_slack_retrievers.py --apply
+
+Flow:
+    1. Find all slack retrievers (category: 'retrievers', type: 'slack')
+    2. Display owner names for each retriever
+    3. Find blueprints that reference these retrievers
+    4. With --apply: delete slack retrievers and associated blueprints
+"""
+
+import os
+import argparse
+from datetime import datetime
+from typing import List, Dict, Any, Set
+import pymongo
+
+
+# ─────────────────────────────────────────────────────────────────────────────
+# Configuration
+# ─────────────────────────────────────────────────────────────────────────────
+
+MONGODB_IP = os.environ.get("MONGODB_IP", "localhost")
+MONGODB_PORT = os.environ.get("MONGODB_PORT", "27017")
+DB_NAME = "UnifAI"
+
+RESOURCES_COLLECTION = "resources"
+BLUEPRINTS_COLLECTION = "blueprints"
+
+# Query for slack retrievers
+SLACK_RETRIEVER_QUERY = {"category": "retrievers", "type": "slack"}
+
+
+# ─────────────────────────────────────────────────────────────────────────────
+# Helper Functions
+# ─────────────────────────────────────────────────────────────────────────────
+
+def find_slack_retrievers(db) -> List[Dict[str, Any]]:
+    """Find all slack retrievers in the resources collection."""
+    collection = db[RESOURCES_COLLECTION]
+    return list(collection.find(SLACK_RETRIEVER_QUERY))
+
+
+def find_blueprints_using_retriever(db, rid: str) -> List[Dict[str, Any]]:
+    """Find blueprints that reference a specific retriever by rid."""
+    collection = db[BLUEPRINTS_COLLECTION]
+    
+    # Blueprints store references in rid_refs array
+    query = {"rid_refs": rid}
+    
+    return list(collection.find(query))
+
+
+def find_blueprints_with_inline_slack(db) -> List[Dict[str, Any]]:
+    """Find blueprints that have inline slack retrievers (not via rid_refs)."""
+    collection = db[BLUEPRINTS_COLLECTION]
+    
+    # Check for inline slack retrievers in spec_dict.retrievers
+    query = {"spec_dict.retrievers.config.type": "slack"}
+    
+    return list(collection.find(query))
+
+
+def delete_retriever(db, rid: str, dry_run: bool = True) -> bool:
+    """Delete a retriever by rid."""
+    if dry_run:
+        return True
+    
+    collection = db[RESOURCES_COLLECTION]
+    result = collection.delete_one({"_id": rid})
+    return result.deleted_count > 0
+
+
+def delete_blueprint(db, blueprint_id: str, dry_run: bool = True) -> bool:
+    """Delete a blueprint by blueprint_id."""
+    if dry_run:
+        return True
+    
+    collection = db[BLUEPRINTS_COLLECTION]
+    result = collection.delete_one({"blueprint_id": blueprint_id})
+    return result.deleted_count > 0
+
+
+# ─────────────────────────────────────────────────────────────────────────────
+# Main Logic
+# ─────────────────────────────────────────────────────────────────────────────
+
+def display_slack_retrievers(retrievers: List[Dict[str, Any]]) -> None:
+    """Display information about slack retrievers."""
+    print(f"\n{'='*60}")
+    print("SLACK RETRIEVERS FOUND")
+    print(f"{'='*60}")
+    print(f"Total: {len(retrievers)}")
+    
+    if not retrievers:
+        print("  No slack retrievers found.")
+        return
+    
+    # Group by owner
+    owners: Dict[str, List[Dict]] = {}
+    for r in retrievers:
+        owner = r.get("user_id", "unknown")
+        if owner not in owners:
+            owners[owner] = []
+        owners[owner].append(r)
+    
+    print(f"\nOwners: {len(owners)}")
+    for owner, items in sorted(owners.items()):
+        print(f"\n  Owner: {owner}")
+        print(f"  {'─'*50}")
+        for item in items:
+            rid = item.get("rid", item.get("_id", "unknown"))
+            name = item.get("name", "unnamed")
+            created = item.get("created", "unknown")
+            cfg = item.get("cfg_dict", {})
+            print(f"    • [{rid}] {name}")
+            print(f"      Created: {created}")
+            print(f"      Config: {cfg}")
+
+
+def display_affected_blueprints(blueprints: List[Dict[str, Any]], 
+                                 retriever_rids: Set[str]) -> None:
+    """Display information about blueprints using slack retrievers."""
+    print(f"\n{'='*60}")
+    print("BLUEPRINTS USING SLACK RETRIEVERS")
+    print(f"{'='*60}")
+    print(f"Total: {len(blueprints)}")
+    
+    if not blueprints:
+        print("  No blueprints using slack retrievers found.")
+        return
+    
+    # Group by owner
+    owners: Dict[str, List[Dict]] = {}
+    for bp in blueprints:
+        owner = bp.get("user_id", "unknown")
+        if owner not in owners:
+            owners[owner] = []
+        owners[owner].append(bp)
+    
+    print(f"\nOwners: {len(owners)}")
+    for owner, items in sorted(owners.items()):
+        print(f"\n  Owner: {owner}")
+        print(f"  {'─'*50}")
+        for bp in items:
+            bp_id = bp.get("blueprint_id", "unknown")
+            spec = bp.get("spec_dict", {})
+            name = spec.get("name", "unnamed")
+            updated = bp.get("updated_at", "unknown")
+            
+            # Find which slack retrievers this blueprint uses
+            rid_refs = bp.get("rid_refs", [])
+            used_slack_refs = [r for r in rid_refs if r in retriever_rids]
+            
+            print(f"    • [{bp_id}] {name}")
+            print(f"      Updated: {updated}")
+            if used_slack_refs:
+                print(f"      Uses retrievers: {used_slack_refs}")
+
+
+def cleanup(db, dry_run: bool = True) -> Dict[str, Any]:
+    """
+    Main cleanup function.
+    
+    Args:
+        db: MongoDB database
+        dry_run: If True, only show what would be deleted
+    
+    Returns:
+        Stats dict with counts
+    """
+    stats = {
+        "retrievers_found": 0,
+        "retrievers_deleted": 0,
+        "blueprints_found": 0,
+        "blueprints_deleted": 0,
+        "errors": []
+    }
+    
+    # Step 1: Find all slack retrievers
+    print("\n🔍 Searching for slack retrievers...")
+    retrievers = find_slack_retrievers(db)
+    stats["retrievers_found"] = len(retrievers)
+    
+    display_slack_retrievers(retrievers)
+    
+    if not retrievers:
+        return stats
+    
+    # Get all retriever rids
+    retriever_rids = set()
+    for r in retrievers:
+        rid = r.get("rid", r.get("_id"))
+        if rid:
+            retriever_rids.add(rid)
+    
+    # Step 2: Find blueprints using these retrievers
+    print("\n🔍 Searching for blueprints using slack retrievers...")
+    
+    all_blueprints: Dict[str, Dict] = {}  # blueprint_id -> blueprint doc
+    
+    # Find via rid_refs
+    for rid in retriever_rids:
+        bps = find_blueprints_using_retriever(db, rid)
+        for bp in bps:
+            bp_id = bp.get("blueprint_id")
+            if bp_id and bp_id not in all_blueprints:
+                all_blueprints[bp_id] = bp
+    
+    # Also find blueprints with inline slack retrievers
+    inline_bps = find_blueprints_with_inline_slack(db)
+    for bp in inline_bps:
+        bp_id = bp.get("blueprint_id")
+        if bp_id and bp_id not in all_blueprints:
+            all_blueprints[bp_id] = bp
+    
+    blueprints = list(all_blueprints.values())
+    stats["blueprints_found"] = len(blueprints)
+    
+    display_affected_blueprints(blueprints, retriever_rids)
+    
+    # Step 3: Delete if not dry run
+    if not dry_run:
+        print(f"\n{'='*60}")
+        print("DELETING RESOURCES")
+        print(f"{'='*60}")
+        
+        # Delete blueprints first (they reference retrievers)
+        print("\n📦 Deleting blueprints...")
+        for bp in blueprints:
+            bp_id = bp.get("blueprint_id")
+            bp_name = bp.get("spec_dict", {}).get("name", "unnamed")
+            try:
+                if delete_blueprint(db, bp_id, dry_run=False):
+                    stats["blueprints_deleted"] += 1
+                    print(f"  ✓ Deleted blueprint: [{bp_id}] {bp_name}")
+                else:
+                    print(f"  ✗ Failed to delete blueprint: [{bp_id}] {bp_name}")
+            except Exception as e:
+                stats["errors"].append(f"Blueprint {bp_id}: {str(e)}")
+                print(f"  ✗ Error deleting blueprint [{bp_id}]: {e}")
+        
+        # Then delete retrievers
+        print("\n🔧 Deleting slack retrievers...")
+        for r in retrievers:
+            rid = r.get("rid", r.get("_id"))
+            name = r.get("name", "unnamed")
+            try:
+                if delete_retriever(db, rid, dry_run=False):
+                    stats["retrievers_deleted"] += 1
+                    print(f"  ✓ Deleted retriever: [{rid}] {name}")
+                else:
+                    print(f"  ✗ Failed to delete retriever: [{rid}] {name}")
+            except Exception as e:
+                stats["errors"].append(f"Retriever {rid}: {str(e)}")
+                print(f"  ✗ Error deleting retriever [{rid}]: {e}")
+    
+    return stats
+
+
+# ─────────────────────────────────────────────────────────────────────────────
+# Main
+# ─────────────────────────────────────────────────────────────────────────────
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Find and clean up Slack retrievers and associated blueprints"
+    )
+    parser.add_argument(
+        "--apply",
+        action="store_true",
+        help="Apply changes - delete slack retrievers and blueprints (default is dry-run)"
+    )
+    parser.add_argument(
+        "--mongodb-ip",
+        default=MONGODB_IP,
+        help=f"MongoDB IP address (default: {MONGODB_IP})"
+    )
+    parser.add_argument(
+        "--mongodb-port",
+        default=MONGODB_PORT,
+        help=f"MongoDB port (default: {MONGODB_PORT})"
+    )
+    args = parser.parse_args()
+    
+    dry_run = not args.apply
+    
+    # Connect to MongoDB
+    mongo_uri = f"mongodb://{args.mongodb_ip}:{args.mongodb_port}/"
+    print(f"Connecting to MongoDB: {mongo_uri}")
+    print(f"Database: {DB_NAME}")
+    
+    client = pymongo.MongoClient(mongo_uri)
+    db = client[DB_NAME]
+    
+    if dry_run:
+        print("\n" + "="*60)
+        print("🔍 DRY RUN MODE - No changes will be made")
+        print("   Use --apply to actually delete resources")
+        print("="*60)
+    else:
+        print("\n" + "="*60)
+        print("⚠️  APPLYING CHANGES - Resources will be deleted!")
+        print("="*60)
+    
+    # Run cleanup
+    stats = cleanup(db, dry_run)
+    
+    # Summary
+    print(f"\n{'='*60}")
+    print("SUMMARY")
+    print(f"{'='*60}")
+    print(f"  Slack retrievers found: {stats['retrievers_found']}")
+    print(f"  Blueprints using slack: {stats['blueprints_found']}")
+    
+    if not dry_run:
+        print(f"\n  Retrievers deleted: {stats['retrievers_deleted']}")
+        print(f"  Blueprints deleted: {stats['blueprints_deleted']}")
+    
+    if stats['errors']:
+        print(f"\n  ❌ Errors: {len(stats['errors'])}")
+        for err in stats['errors']:
+            print(f"    - {err}")
+    
+    if dry_run:
+        print(f"\n📌 This was a DRY RUN. Use --apply to delete resources.")
+    else:
+        print(f"\n✅ Cleanup complete!")
+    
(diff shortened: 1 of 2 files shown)


backend/celery_app/tasks/pipeline_tasks.py
```diff
@@ -63,4 +67,9 @@ def execute_pipeline_task(self, source_type: str, source_data: dict):
...
         
     except Exception as e:
-        logger.error(f"Pipeline execution failed for {source_type}: {str(e)}", exc_info=True)
\ No newline at end of file
+        logger.error(f"Pipeline execution failed for {source_type}: {str(e)}", exc_info=True)
+    
+    finally:
+        # Clean up uploaded file after pipeline completes (success or failure) for documents
+        if source_type.upper() == DataSource.DOCUMENT.upper_name and doc_path:
+            cleanup_file(doc_path, "after pipeline completion")
```
Nirsisr: @MayaCrmi I would prefer if we can handle operations such cleanup from the pipelineFactory, so for each SOURCE we will have it's own implementation for cleanup. It's ok if we don't have such for slack so Just return in the .cleanup() call over there 'NotImplemented' and from docs you call the relevant cleanup.

backend/data_sources/docs/doc_config_manager.py
```diff
@@ -15,7 +15,7 @@ class DocConfigManager(ConfigurationManager):
         "include_metadata": True,
         "chunk_size": 1000,
         "chunk_overlap": 200,
-        "supported_extensions": [".pdf", ".docx", ".txt", ".md"],
+        "supported_extensions": [".pdf", ".docx", ".md", ".pptx"],
```
Nirsisr: @yhabushi79 once we move to docling as a service, make sure you are synced here.

ui/client/src/features/docs/UploadTab.tsx
```diff
@@ -60,20 +80,24 @@ export const UploadTab: React.FC<UploadTabProps> = ({
...
     };
-    
-    const validateFiles = (files: FileList): { validFiles: File[], invalidFiles: string[] } => {
-        const validFiles: File[] = [];
-        const invalidFiles: string[] = [];
+
+    /**
+     * Checks if a file with the same normalized name already exists for the current user
+     */
+    const checkDuplicateFileName = (fileName: string): { isDuplicate: boolean; normalizedName: string; existingDoc?: Document } => {
```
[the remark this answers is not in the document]
MayaCrmi: Of course. While thinking on it now I understand a small problem that if a pipeline failed I do want to allow the user to upload the same file... will fix it now :)

backend/celery_app/tasks/pipeline_tasks.py
MayaCrmi: ok creating cleanup function on the pipeline and on slack specifically with NotImplemented, and using the doc cleanup fucntion in the docs_pipeline

backend/registration/document.py
```diff
@@ -56,6 +60,9 @@ def run_validator(self) -> Tuple[bool, Dict[str, Any] | None]:
         is_valid, issue = self._validator.validate(**validation_args)
 
         if not is_valid:
+            # Clean up the uploaded file since it failed validation
+            cleanup_file(self.source_data.doc_path, "after validation failure")
```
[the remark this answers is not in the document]
MayaCrmi: this now calls the cleanup_file from global utils

backend/pipeline/pipeline.py
```diff
@@ -72,4 +72,8 @@ def store_embeddings(self, embeddings: List[Dict]):
         return self.storage.store_embeddings(embeddings)
     
     def clean_orchestration(self):
-        self.monitor.finish_log_monitoring()
\ No newline at end of file
+        self.monitor.finish_log_monitoring()
+    
+    def cleanup(self):
```
Lina-AbuYousef: If this method isn’t abstract and doesn’t contain any logic in the base class, since the logic is specific to docs—why do we need to keep it in the base class?

backend/registration/document.py
```diff
@@ -32,14 +32,17 @@ def __init__(self, mongo_storage: Any, upload_by: str, instance: Dict[str, Any])
 
     @cached_property
     def source_data(self) -> DocumentSourceData:
-        name = self.instance.get("source_name", "")
-        path = os.path.join(self.upload_folder, name)
+        # Get the original name for display purposes
+        original_name = self.instance.get("source_name", "")
+        # Use secure_filename to get the actual filename which matches what upload_docs() does when saving the file
+        secure_name = secure_filename(original_name)
```
Lina-AbuYousef: so this line that handle the case where we might have two uploaded files with the same name?

Lina-AbuYousef: ooh its not related, ignore pls

ui/client/src/features/docs/UploadTab.tsx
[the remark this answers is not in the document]
MayaCrmi: I did (:

backend/registration/document.py
MayaCrmi: NPP Resolving

MayaCrmi:
Explaining changes I made now in data_sources.py:
I found an error occurring when using the bulk delete. In bulk delete, we are doing this multiple times in the same process:
storage_manager = initialize_storage_manager(...)
So during bulk delete the first delete loads the model fine, and the second delete is another model load while the previous one hasn't been fully freed etc. To fix it, instead of creating an embedding for each one I just send now the default embedding_dim to vector stroage for the deletion case (added a for_delete bool to the initilization so this will only be done when deleting in case we need it for other stuff though for now its only called from delete.
If you think this is a problem I can also instead add a test like so:
```
_embedding_generator = None

def initialize_embedding_generator():
    global _embedding_generator
    if _embedding_generator is None:
        embedding_config = asdict(EmbeddingConfig())
        _embedding_generator = EmbeddingGeneratorFactory.create(embedding_config)
    return _embedding_generator
```
But I didn't feel like we need this for here.

ui/client/src/components/shared/DataTable.tsx
```diff
@@ -150,16 +171,86 @@ export function DataTable<T extends object>({
...
+        const filteredRows = table.getFilteredRowModel().rows
+        const isAllFilteredSelected = filteredRows.length > 0 && filteredRows.every(row => {
+          const rowId = getRowId ? getRowId(row.original) : row.id
+          return rowSelection[rowId]
+        })
+        
+        return (
+          <Checkbox
+            checked={isAllFilteredSelected}
+            onCheckedChange={(checked) => {
```
Lina-AbuYousef: Its better to have a separated function that do this logic and call it here, it will be more readable

ui/client/src/components/shared/DataTable.tsx
```diff
@@ -150,16 +171,86 @@ export function DataTable<T extends object>({
...
+            aria-label="Select all filtered rows"
+          />
+        )
+      },
+      cell: ({ row }) => {
+        const rowId = getRowId ? getRowId(row.original) : row.id
+        return (
+          <Checkbox
+            checked={rowSelection[rowId] === true}
+            onCheckedChange={(checked) => {
```
Lina-AbuYousef: same

ui/client/src/components/shared/DataTable.tsx
```diff
@@ -181,13 +272,18 @@ export function DataTable<T extends object>({
       }
     },
     onGlobalFilterChange: setGlobalFilter,
-    onColumnFiltersChange: setColumnFilters,
+    onColumnFiltersChange: (updater) => {
+      const newFilters = typeof updater === 'function' ? updater(columnFilters) : updater
```
Lina-AbuYousef: I'm not sure how is the filters related to the new checkbox column, could u pls explain here

ui/client/src/features/docs/DocumentsPage.tsx
```diff
@@ -22,24 +28,59 @@ export default function Documents() {
...
+      return hasActive ? 5000 : false;
+    },
   });
 
   useEffect(() => {
     resetPage();
   }, []);
 
+  // Refetch documents immediately when upload modal closes to show new documents
+  useEffect(() => {
```
Lina-AbuYousef: niceeee

ui/client/src/features/docs/UploadTab.tsx
```diff
@@ -32,25 +37,33 @@ interface FileWithTags {
...
     const fileInputRef = useRef<HTMLInputElement>(null);
 
+    // Fetch existing documents to check for duplicates
+    const { data: existingDocuments = [] } = useQuery<Document[]>({
+        queryKey: ['documents'],
+        queryFn: fetchDocuments,
+        staleTime: 30 * 1000,
+    });
+
     useEffect(() => {
```
Lina-AbuYousef: did you wanted this component to be kind of standalone so u did the fetch extensions here again ? just wondering since it may be a good approach and reason

Lina-AbuYousef: also, If you have used the queryClient i think you can just used here the cached extensions form it instead from the API

Lina-AbuYousef: Like this cached is very useful when we are using the same API that return also static result across many places in the UI

ui/client/src/components/shared/DataCard.tsx
```diff
@@ -1,10 +1,10 @@
 import { useState } from "react";
```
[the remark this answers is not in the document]
MayaCrmi: components/ui/ contains the low-level UI components (button, checkboxes, etc) which we got when we first created the app, while shared/  are more higher-level, reusable components that use the ones in ui and add feature-specific logic, which are the ones that we created later on during development

ui/client/src/components/shared/DataTable.tsx
```diff
@@ -95,13 +98,19 @@ interface DataTableProps<T extends object> {
   enableGlobalFilter?: boolean
   enableColumnFilters?: boolean
   enablePagination?: boolean
+  enableRowSelection?: boolean
```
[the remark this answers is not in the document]
MayaCrmi: enableRowSelection is used to enable/disable the row selection feature (shows/hides the checkbox column) and rowSelection is the state object tracking which rows are currently selected

ui/client/src/components/shared/DataTable.tsx
```diff
@@ -133,8 +147,15 @@ export function DataTable<T extends object>({
     pageSize: initialState?.pagination?.pageSize ?? 10,
   })
 
+  const [internalRowSelection, setInternalRowSelection] = React.useState<RowSelectionState>(
+    initialState?.rowSelection ?? {}
+  )
+  
+  const rowSelection = controlledRowSelection ?? internalRowSelection
```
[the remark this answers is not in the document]
MayaCrmi: controlled mode means the parent passes  the data, in this case rowSelection + onRowSelectionChange so DataTable uses those (external state management)
while uncontrolled mode means DataTable manages its own internal state
so here DocumentsTable uses controlled mode to share selection state with the bulk delete button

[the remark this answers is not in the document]
Lina-AbuYousef: Aha so its like customize functionality

Nirsisr:
> Explaining changes I made now in data_sources.py: I found an error occurring when using the bulk delete. In bulk delete, we are doing this multiple times in the same process: storage_manager = initialize_storage_manager(...) So during bulk delete the first delete loads the model fine, and the second delete is another model load while the previous one hasn't been fully freed etc. To fix it, instead of creating an embedding for each one I just send now the default embedding_dim to vector stroage for the deletion case (added a for_delete bool to the initilization so this will only be done when deleting in case we need it for other stuff though for now its only called from delete. If you think this is a problem I can also instead add a test like so:
> 
> ```
> _embedding_generator = None
> 
> def initialize_embedding_generator():
>     global _embedding_generator
>     if _embedding_generator is None:
>         embedding_config = asdict(EmbeddingConfig())
>         _embedding_generator = EmbeddingGeneratorFactory.create(embedding_config)
>     return _embedding_generator
> ```
> 
> But I didn't feel like we need this for here.

@MayaCrmi as we discussed, this should be saved once we will move the Embedding Model under a separated service to handle with that, same for the doclinc. My question is why we need the EmbeddingGenerator for deletion? 

Something of the core logic doesn't feel right at all .. 
We need an accesses to Qdrant for deletion. No a single connection to the Embedding mode.. Please explain

ui/client/src/components/shared/RowSelectionCheckbox.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { RowSelectionState } from "@tanstack/react-table";
+
+interface RowSelectionCheckboxProps {
+  rowId: string;
+  rowSelection: RowSelectionState;
+  onRowSelectionChange: (selection: RowSelectionState) => void;
+  ariaLabel?: string;
+}
+
+export function RowSelectionCheckbox({
```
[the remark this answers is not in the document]
Nirsisr: Good catch here

ui/client/src/utils/documentValidation.ts
```diff
@@ -0,0 +1,106 @@
...
+export interface SelectedFileItem {
+  file: File;
+  id: string;
+}
+
+/**
+ * Filters out duplicate files from a list of valid files.
+ * Returns non-duplicate files and a list of duplicate files with their info.
+ */
+export const filterDuplicateFiles = (
```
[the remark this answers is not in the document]
Nirsisr: @MayaCrmi I agree with Lina here, first we must have it on the backend (along the GUI or only from there) to support those actions being executed from an API level.

Second, when calling to /upload API (not sure about the naming) first thing to do is the VALIDATION layer, if something wrong, it will raise an ERROR to the GUI with the data like: `Duplicated file name` and GUI will throw a toast, so in that case UPLOAD not begin as well, validations should be handled from the BE as a rule of thumb, it's not good we have it from the GUI.

Regard validation such as FILE_SIZE this one is exception of the RULE OF THUMB, why? Since we want block the API call with a huge file (that need to be sene by a File descriptor between GUI to Backend from even happen) that's why here we are kind of blocking it from GUI side to prevent heavy API call. (but still we need validation layer for this error as well on the backend, kind of protecting it if someone use our API directly)

----------------------------------------------------------------------------------------------------------------
Regard API support in general we need to be aware of that since it's on the Roadmap also for RAG areas.

backend/providers/data_sources.py
```diff
@@ -35,10 +35,14 @@ def initialize_vector_storage(embedding_dim: int = 384, source_type: str = "data
...
+def initialize_storage_manager(source_type: str = "data_source", for_delete: bool = False):
     """Initialize and return storage manager for operations."""
-    embedding_generator = initialize_embedding_generator()
-    vector_storage = initialize_vector_storage(embedding_generator.embedding_dim, source_type)
+    if for_delete:
+        embedding_dim = 384  # Assuming fixed embedding dim for deletion
+    else:
+        embedding_generator = initialize_embedding_generator()
+        embedding_dim = embedding_generator.embedding_dim
+    vector_storage = initialize_vector_storage(embedding_dim=embedding_dim, source_type=source_type)
```
Nirsisr: We discussed F2F about Investigation for the `Singleton Creation` (SentenceTransformerEmbedding) called by `EmbeddingGeneratorFactory` in a sequence one after another.

Also need to call /delete for [id: string] instead of calling the API 1 after another, which makes unnecessary overhead.

MayaCrmi:
### Major changes have been pushed:
* Now the entire validation process happens in a seperate API that runs when the user uploads the files from his computer (before submission). It it very well-documented in the code itself so I will not explain too much here, let me know if you want me to elaborate.
* Added GENIE-1069 here (not sending the full_text when getting the list of sources, but only on demand when a user clicks the eye icon)
* Bulk delete logic has been improved, for that we needed to change existing logic for the delete apis to fit the new option.
* Improvements in pagination logic and location, including a problem in the footer.

backend/endpoints/data_sources.py
```diff
@@ -23,32 +24,90 @@ def available_data_sources(source_type):
...
+            top_level_result = results["succeeded"][0].get("result", {})
+        
+        if len(results["failed"]) == 0:
+            response = {
+                "status": "success",
+                "message": f"Successfully deleted {len(results['succeeded'])} source(s)",
+                "results": results
+            }
+            if top_level_result:
+                response["result"] = top_level_result
```
Nirsisr: @MayaCrmi I don't like that, why we need any backward compatability?

If we have another data_source == SLACK :) that leveraging it, we need to change the way it calls the API and gets the payload with our current implementation.

It's not elegant/looking good we keep code snippet for backward compatibility like we MUST avoid it if we can.

backend/endpoints/docs.py
```diff
@@ -24,6 +25,64 @@ def upload_files(files):
     except Exception as e:
         logger.error(f"Failed to upload files: {str(e)}")
         return jsonify({"error": str(e)}), 500
+
+
+@docs_bp.route("/validate", methods=["POST"])
```
Nirsisr: That's code make my SOULD happy, exactly what we want, professional !

backend/endpoints/pipelines.py
```diff
@@ -23,6 +24,15 @@ def start_pipeline(data, source_type, logged_in_user):
         data: List of data sources to register and process
         source_type: Type of data source (SLACK, DOCUMENT, etc.)
         logged_in_user: Username of the current user
+        skip_validation: If True, skip file validation during registration.
+            This should ONLY be True when files have been pre-validated via 
+            the /docs/validate endpoint (UI flow).
+            
+            For external API calls (Postman, scripts), this should be False (default)
```
Nirsisr: Maya's plead remove Postman here : D

Maybe I am קטנוני but it's clear saying external API calls like POSTMAN we love using in the team but no need to mention it explicitly in the code.

backend/services/documents/file_validation_service.py
Nirsisr: @MayaCrmi 10x for putting well documentation for this as requested, appreciate it.

When I just read it, I understand the entire flow that's why it so crucial when things won't be fresh in our heads, we know exactly what we cover, that's perfect.

backend/endpoints/pipelines.py
MayaCrmi: haha of course! Done

backend/services/documents/file_validation_service.py
```diff
@@ -0,0 +1,324 @@
...
+        
+        logger.info(
+            f"File validation complete for user {self.username}: "
+            f"{len(valid_files)} valid, {len(errors)} errors"
+        )
+        
+        return FileValidationResult(valid_files=valid_files, errors=errors)
+
+
+def validate_files_for_user(
```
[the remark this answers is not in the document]
MayaCrmi: NP, the method name is now validate()  and I removed after your previous comment as well the validate_files_for_user() method entiorely so the endpoint now uses the class directly like initializes the service. I agree that it alignes better with what was already implemented.

backend/endpoints/data_sources.py
```diff
@@ -23,32 +24,90 @@ def available_data_sources(source_type):
     
 
 @data_sources_bp.route("/data.source.delete", methods=["DELETE"])
-@from_body({"pipeline_id": fields.Str(required=True)})
-def delete_source(pipeline_id):
+@from_body({"pipeline_ids": fields.List(fields.Str(), required=True)})
+def delete_source(pipeline_ids):
```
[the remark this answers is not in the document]
MayaCrmi: yes slack is now also sending deletion with list format for the new structure, but since we changed things in your next comment I solved a problem there now (backward compability)

backend/providers/data_sources.py
```diff
@@ -132,4 +136,41 @@ def get_data_source_by_id(pipeline_id: str, source_type: str = None):
...
+
+        if not result.get("success"):
+            return result
+
+        source_info = result.get("source_info", {})
+        pipeline_id = result.get("pipeline_id")
+
+        # Enrich with pipeline stats if available
+        if pipeline_id:
+            pipeline_stats = svc.get_pipeline_stats([pipeline_id])
```
Lina-AbuYousef: As I remember the get_available_data_sources above is returning the list of data sources enriched already with the stats, I'm attaching it here : 
```
 def list_sources(self, source_type: Optional[str] = None) -> List[Dict[str, Any]]:
        """Get all sources enriched with pipeline stats (for backward compatibility)."""
        sources = self.sources.get_all(source_type)
        pipeline_ids = [s.get('pipeline_id') for s in sources if s.get('pipeline_id')]
        valid_ids = [pid for pid in pipeline_ids if pid is not None]
        pipeline_stats = self.pipelines.get_stats(valid_ids)
        
        enriched = []
        for source in sources:
            pipeline_id = source.get('pipeline_id')
            if pipeline_id and pipeline_id in pipeline_stats:
                source['pipeline_stats'] = pipeline_stats[pipeline_id]
                source['status'] = pipeline_stats[pipeline_id].get('status')
            else:
                source['pipeline_stats'] = None
                source['status'] = None
            enriched.append(make_json_safe(source))
        
        enriched_sorted = sorted(
            enriched,
            key=lambda s: s.get('created_at') or 0,  # default to 0 if created_at is missing
            reverse=True
        )
        return enriched_sorted
```

So if this function created only for enriching with stats then you can use the get_available_data_sources one

backend/registration/document.py
```diff
@@ -22,24 +22,42 @@ class DocumentSourceData(BaseSourceData):
     md5: str
     
 class DocumentRegistration(RegistrationBase):
-    """Registration flow for Document sources."""
+    """
+    Registration flow for Document sources.
+    
+    Supports skip_validation flag:
+    - When False (default): Full validation (extension, size, name, MD5)
```
Lina-AbuYousef: So now in the UI the user will already get the validation result int he upload page right

backend/services/documents/duplicate_checker.py
```diff
@@ -10,18 +14,53 @@ class DocumentDuplicateChecker:
...
+    A document is only considered a duplicate if an existing document with the 
+    same MD5 hash has been successfully processed (status = DONE). Documents 
+    with FAILED or other non-complete statuses are not considered duplicates,
+    allowing users to retry failed uploads.
     """
 
     def __init__(self, storage: Any):
         self._storage = storage
 
+    def _get_pipeline_status(self, pipeline_id: str) -> Optional[str]:
```
Lina-AbuYousef: Could we move this function out of this class?
This class is meant to handle validation using MD5 only. as better to keep on the single responsibility and clean classes 
If you need the status, we can use get_stats from PipelinesRepository, since it already returns the status, or pass the status into this class instead.

backend/services/documents/duplicate_checker.py
```diff
@@ -34,7 +73,20 @@ def find_existing_by_md5(
...
             result = self._storage.get_source_by_query(query)
-            return result if isinstance(result, list) else []
+            sources = result if isinstance(result, list) else []
+            
+            if not only_blocking or not sources:
+                return sources
+            
+            # Filter to only include sources with blocking status (DONE)
+            blocking_sources = []
+            for source in sources:
```
Lina-AbuYousef: Also, why are we skipping the other documents here "not DONE"? I’m not sure this is correct.
For example, if we have a document that’s currently running and another document with the same name and MD5, how do we detect that case?
Wouldn’t this potentially create a bug where a newly uploaded file with the same content is missed because another one is already running?
maybe I didn't get the purpose of this lines

ui/client/src/api/docs.ts
```diff
@@ -61,9 +66,82 @@ export async function getSupportedFileExtensions(): Promise<string[]> {
...
+}
+
+/**
+ * Response from the file validation endpoint
+ */
+export interface FileValidationResponse {
+    valid_files: ValidatedFile[];
+    errors: FileValidationError[];
+    has_errors: boolean;
+}
```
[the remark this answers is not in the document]
Nirsisr: Of course I mean docs.ts : D

backend/registration/document.py
MayaCrmi: yes (:

backend/utils/storage/mongo/sources_repository.py
```diff
@@ -11,10 +11,15 @@ class SourcesRepository:
...
+    def get_all(self, source_type: Optional[str] = None, projection: Optional[Dict[str, Any]] = None) -> List[Dict[str, Any]]:
+        """Get all sources, optionally filtered by type.
+        
+        Args:
+            source_type: Optional filter by source type
+            projection: Optional MongoDB projection to include/exclude fields
+        """
         query = {"source_type": source_type.upper()} if source_type else {}
-        return list(self.col.find(query))
+        return list(self.col.find(query, projection))
```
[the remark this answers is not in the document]
MayaCrmi: :-*

backend/services/documents/duplicate_checker.py
MayaCrmi: yes done

backend/services/documents/file_validation_service.py
```diff
@@ -0,0 +1,220 @@
...
+                    "error_type": e.error_type,
+                    "message": e.message
+                }
+                for e in self.errors
+            ],
+            "has_errors": len(self.errors) > 0
+        }
+
+
+class FileValidationService:
```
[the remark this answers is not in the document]
MayaCrmi: the discussion was on whether the new service is necessary given that the existing layer already manages running individual DataSourceValidator classes, but this service was introduced to support broader validation scenarios not currently covered by the base layer (pre-upload validation) and returning UI-friendly responses. The one already existing is designed for single-item, post-registration flows. Nir proposed aligning both approaches into a single validation layer that supports PRE and POST stages, allowing different validator sets to be initialized based on the stage. At the end it was agreed that this refactor is important but too large for the current PR which already went through many cycles and has like 50 changed files, so I have a ticket for it which I'll work on as soon as this is merged

odaiodeh (approved):

PR #32, opened 2025-12-17
Genie 1072/cr agent be

sfiresht:
This PR adds CR agent review to the backend code.

docs(backend): add architecture guide and revamp readme

- Add detailed ARCHITECTURE.md covering patterns, pipelines, and conventions
- Update README.md with comprehensive setup and deployment instructions
- Remove obsolete documentation

### DIFF
--- a/ui/ARCHITECTURE.md
+++ b/ui/ARCHITECTURE.md
@@ -956,7 +956,7 @@ export async function fetchActivePipelines(): Promise<ActivePipeline[]> {
 |--------|--------|----------|----------|
 | `api` | `@/http/queryClient` | `/api1` (Data Pipeline) | Document/Slack pipelines |
 | `axios` | `@/http/axiosAgentConfig` | `/api2` (Multi-Agent) | Agentic workflows, sessions |
-| _(default)_ | `@/http/authClient` | `/api3` (SSO) | Auth, user management |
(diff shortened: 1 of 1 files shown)


Nirsisr (dismissed):
@sfiresht 10x for it saariko. I approved, but I **want & need** to go over the README + ARCHI file, maybe we do it togeher in the office with @Lina-AbuYousef @MayaCrmi to make sure we are all agree with each little word written there.

That's MUST be perfect so we have valuable CR insights, 
10x saariko !

yhabushi79 (dismissed):

ui/ARCHITECTURE.md
Nirsisr: @sfiresht You told me, you catch an error over here? What`s the error? 

I don't see you change anything, you just change the order of lines or am I missing something? And you change the Client to `_default_` instead of `api` and I'm not sure why

sfiresht: see below, api1 and api3 where replced for some reason. it was caught when I reviewed the umami actually. I kept getting that there's a problem in the doc against the code.

Nirsisr: Saar you confused me, I go over @ARCHITECTURE.MD file and all is right there (I am looking on main), like api1-api3 are all configured as we are using them.

Nirsisr (approved):

PR #39, opened 2025-12-24
GENIE-1076/story/display-names-instead-of-ids

MayaCrmi:
### Overview

This change modifies how document references are displayed and stored in the DocsDataflowRetriever configuration. Instead of displaying document IDs to users, the UI now shows meaningful document names while storing full objects (with both id and name) in the database.

Before:

image

After:

image

### Solution

Data Format Change:

Before: doc_ids: ["abc123", "def456"]

After:  doc_ids: [{"id": "abc123", "name": "Project Proposal"}, {"id": "def456", "name": "Technical Specification"}]

The UI displays names; the backend extracts IDs for Dataflow queries.

---

#### Multi-agent Backend

|File|Change|
|---|---|
|core/field_hints.py|Renamed label_field → display_field; updated description to document dot-notation path support; added display_name to the interface to display "documents" instead of "doc_ids"|
|elements/retrievers/docs_dataflow/config.py|Changed doc_ids type from List[str] to List[Union[dict, str]]; chanegd to display_field="name"; updated description to the new structure; added display_name to the field's configuration|
|elements/retrievers/docs_dataflow/docs_dataflow_retriever.py|Added _extract_values() helper to extract IDs from mixed array (supports both old string format and new object format) to be used in the actual retriever part|

#### Frontend

|File|Change|
|---|---|
|ui/client/src/utils/displayUtils.ts|NEW - Shared utilities: getDisplayValue(), getDisplayValueFromItem(), simplifyConfigForDisplay()|
|ui/client/src/components/agentic-ai/workspace/fieldPopulationUtils.ts|NEW - Population utilities: OptionItem interface, normalizeOptions(), resolvePath(), extractDisplayName(), extractId()|
|ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx|Updated to store full objects via originalObject; displays names via getDisplayLabel(); sends objects to parent for backend storage|
|ui/client/src/components/agentic-ai/workspace/FieldRenderer.tsx|Added isArrayWithPopulate check for anyOf schemas; extracts names for textarea display via getItemLabel() with dot-notation support|
|ui/client/src/components/agentic-ai/workspace/ElementData.tsx|Uses simplifyConfigForDisplay() for config display|
|ui/client/src/components/agentic-ai/workspace/ElementGrid.tsx|Uses getDisplayValueFromItem() for card display|
|ui/client/src/components/agentic-ai/workspace/ElementForm.tsx|Updated to support the new structure|
|ui/client/src/workspace/ResourceDetailsModal.tsx|Uses simplifyConfigForDisplay() for config display|

#### Migration

|File|Description|
|---|---|
|scripts/migrate_doc_ids.py|Migrates existing resources from string array to object array format|

---

### Data Flow

┌─────────────────────────────────────────────────────────────────────┐

│                         FieldPopulation                              │

│  - Fetches from API: [{id, name}, ...]                              │

│  - Normalizes to OptionItem: {label, value, originalObject}        │

│  - Sends originalObject[] to parent via onPopulateResult           │

└───────────────────────────────┬─────────────────────────────────────┘

                                │

                                ▼

┌─────────────────────────────────────────────────────────────────────┐

│                          ElementForm                                 │

│  - Stores full objects in formData[fieldName]                       │

│  - Submits to backend as [{id, name}, ...]                          │

└───────────────────────────────┬─────────────────────────────────────┘

                                │

                                ▼

┌─────────────────────────────────────────────────────────────────────┐

│                         FieldRenderer                                │

│  - Reads value (array of objects)                                   │

│  - Extracts names via getItemLabel() using display_field            │

│  - Displays comma-separated names in Textarea                       │

└─────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────┐

│                       Backend (Retriever)                            │

│  - Receives [{id, name}, ...]                                       │

│  - Extracts IDs via _extract_values()                               │

│  - Uses IDs for Dataflow queries                                    │

└─────────────────────────────────────────────────────────────────────┘

---

### Key Technical Details

1. Backwards Compatibility: Both old (["id1", "id2"]) and new ([{id, name}, ...]) formats are supported via Union[dict, str] type and _extract_values() helper.

2. anyOf Schema Support: FieldRenderer now checks for anyOf containing array types, handling Optional[List[...]] schemas correctly.

3. Dot-notation Path Support: The display_field supports paths like "name.x" for nested objects (e.g., document.name.x).

4. Utility Consolidation: Display logic is centralized in:

- displayUtils.ts - Generic display helpers

- fieldPopulationUtils.ts - Population-specific helpers

---

### Migration

Run the migration script after deployment:

# Dry run first

python scripts/migrate_doc_retrievers.py --dry-run

# Apply migration

python scripts/migrate_doc_retrievers.py

See README_migrate_doc_retrievers.md for full documentation.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Enhanced field configuration with improved display and naming support for better field selection in forms.
  * Extended document identifier support to accept both object-based and string identifiers for greater flexibility.

* **Improvements**
  * Refined multi-select form field handling and value normalization.
  * Enhanced configuration value display formatting for clearer information presentation.
  * Improved array field rendering with better placeholder and label logic.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/multi-agent/elements/retrievers/docs_dataflow/config.py
+++ b/multi-agent/elements/retrievers/docs_dataflow/config.py
@@ -1,4 +1,4 @@
-from typing import Literal, Optional, List, Union
+from typing import Dict, List, Literal, Optional
 from pydantic import Field
 from elements.retrievers.common.base_config import BaseRetrieverConfig
 from core.field_hints import ActionHint, HintType, SelectionType
@@ -31,7 +31,7 @@ class DocsDataflowRetrieverConfig(BaseRetrieverConfig):
         description="Request timeout in seconds"
     )
 
-    docs: Optional[List[Union[dict, str]]] = Field(
+    docs: Optional[List[Dict]] = Field(
         default=None,
         description="Filter results to specific documents",
         json_schema_extra=ActionHint(
--- a/multi-agent/elements/retrievers/docs_dataflow/docs_dataflow_retriever.py
+++ b/multi-agent/elements/retrievers/docs_dataflow/docs_dataflow_retriever.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Union
+from typing import Dict, List, Optional
 from elements.retrievers.common.base_retriever import BaseRetriever
 from elements.providers.dataflow_client.config import DataflowProviderConfig
 from elements.providers.dataflow_client.dataflow_provider_factory import DataflowProviderFactory
@@ -15,7 +15,7 @@ class DocsDataflowRetriever(BaseRetriever):
             top_k_results: int,
             threshold: float,
             timeout: float = 30.0,
-            docs: Optional[List[Union[dict, str]]] = None,
+            docs: Optional[List[Dict]] = None,
             tags: Optional[List[str]] = None,
     ):
         self.threshold = threshold
@@ -31,8 +31,8 @@ class DocsDataflowRetriever(BaseRetriever):
     def retrieve(self, query: str) -> List[dict]:
         context = get_current_context()
 
-        # Extract document IDs from docs list (handles both dict and string items)
-        doc_ids = [doc.get('id') if isinstance(doc, dict) else doc for doc in (self.docs or [])] or None
+        # Extract document IDs from docs list
+        doc_ids = [doc['id'] for doc in self.docs] if self.docs else None
 
         response = self._provider.query(
             query=query,


ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
```diff
@@ -117,45 +111,60 @@ export const FieldPopulation: React.FC<FieldPopulationProps> = ({
...
-      if (typeof item === 'object' && 'label' in item && 'value' in item) {
-        return {
-          label: String(item.label),
-          value: String(item.value)
-        };
+      // Fallback to formData for items not in populated options
+      const currentValue = formData[fieldName];
+      if (Array.isArray(currentValue)) {
+        const item = currentValue.find((i: any) => extractValue(i) === value);
+        if (item) return item;
```
Nirsisr: How can this fallback appear? Currently we are under fieldpopulation object so why we won't see the items in the populated options, what am I missing?

ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
Nirsisr: Maya I see here a lot of changes related to previos feature of ALL Select, why is it here again? This one is up-to-date with main?

ui/client/src/components/agentic-ai/workspace/FieldRenderer.tsx
Nirsisr: Let's discuss it on 1 ON 1, not sure I completely supportive of the changes here, we need to re-think

ui/client/src/components/agentic-ai/workspace/fieldPopulationUtils.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (displayField) {
+    const val = resolvePath(obj, displayField);
+    if (val != null) return String(val);
+  }
+
+  // Common display field fallbacks
+  const fallbacks = ['name', 'label', 'title', 'display', 'text'];
+  for (const key of fallbacks) {
+    if (obj[key] != null) return String(obj[key]);
+  }
```
Nirsisr: I don't like those fallbacks, it's against to BE-GUI protocols. Violation of the populateHint contract.

We need to come up with something better

MayaCrmi: made some changes (response for this and the rest of the comments here)

multi-agent/elements/retrievers/docs_dataflow/docs_dataflow_retriever.py
```diff
@@ -10,16 +10,38 @@ class DocsDataflowRetriever(BaseRetriever):
...
+        Items can be dicts (with value stored in value_field) or plain strings.
+        """
+        if not items:
+            return None
+        
+        result = []
+        for item in items:
+            if isinstance(item, dict):
+                # Extract value from dict using the value_field key
+                value = item.get(value_field) or item.get('value')
```
Lina-AbuYousef: Not sure what item.get('value') should return , could u explain this line

ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
```diff
@@ -355,17 +359,17 @@ export const FieldPopulation: React.FC<FieldPopulationProps> = ({
...
         onPopulateResult(fieldName, rawResults, false);
         return;
       }
       
-      const normalizedResults = normalizeOptions(rawResults);
+      const normalizedResults = normalizeOptions(rawResults, displayField, valueField);
       const newOptionValues = new Set(normalizedResults.map(opt => opt.value));
       
-      if (populateHint.selection_type == 'manual' || populateHint.multi_select) {
+      if (populateHint.selection_type === 'manual' || populateHint.multi_select) {
```
Nirsisr: @MayaCrmi No need for the triple === equal sign here (we used it when we want to check ref of objects in the memory) like no need use it for string comparisons

ui/client/src/components/agentic-ai/workspace/FieldRenderer.tsx
```diff
@@ -498,6 +499,100 @@ export const FieldRenderer: React.FC<FieldRendererProps> = ({
       </div>
     );
   }
+  
+  // Handle array fields WITH populate hint - show textarea with names + FieldPopulation for selection
+  // Also handles anyOf schemas (from Optional[List[...]]) that contain an array type
+  const isArrayWithPopulate = populateHint && (
```
Nirsisr: Discussed F2F, please remove any dependency of `populateHint` over here, render methodology should remain the same. 

Tackle the case when it's array, and the field is `populateHint` from the inner renders of the <text> component (all the details where F2F)

ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
MayaCrmi: removed

ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
```diff
@@ -73,12 +66,11 @@ export const FieldPopulation: React.FC<FieldPopulationProps> = ({
   const [isLoading, setIsLoading] = useState(false);
   const [isLoadingMore, setIsLoadingMore] = useState(false);
   const [populatedOptions, setPopulatedOptions] = useState<OptionItem[]>([]);
-  const [selectedValues, setSelectedValues] = useState<string[]>(currentValue);
+  const [selectedValues, setSelectedValues] = useState<string[]>([]); // Track by value (ID)
```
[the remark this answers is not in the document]
MayaCrmi: removed

ui/client/src/components/agentic-ai/workspace/arrayFieldHelpers.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (value.length === 0) return '';
+    return value.map(item => getItemLabel(item, displayFieldPath)).join(', ');
+  }
+  return getItemLabel(value, displayFieldPath);
+};
+
+/**
+ * Array field mode types
+ */
+export type ArrayFieldMode = 'refItems' | 'populate' | 'regular';
```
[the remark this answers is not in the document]
Nirsisr: The naming should be `dynamic` (since the return value might change)

multi-agent/elements/retrievers/docs_dataflow/config.py
```diff
@@ -31,15 +31,16 @@ class DocsDataflowRetrieverConfig(BaseRetrieverConfig):
...
     )
 
-    doc_ids: Optional[List[str]] = Field(
+    doc_ids: Optional[List[Union[dict, str]]] = Field(
         default=None,
-        description="Filter results to specific document IDs",
+        description="Filter results to specific documents",
         json_schema_extra=ActionHint(
             action_uid="dataflow.get_available_docs",
+            display_name="documents",
```
odaiodeh: @MayaCrmi  what is this display name exactly?

MayaCrmi: you requested an example for this part so here's an explanation on what which field here means (shortly for ones you created and more explained for the changes I made):
so now the ui receives the action if for this and it will call the action and receive a list of documents which is now built like this: [{"id": "1", "name": "a"}, {"id": "2", "name": "b"}]. 
then I put display_name="documents" which is the label shown in the UI for the population action since before it was displayed as doc_ids but now we don't show them ids but names so it makes more sense to be called that.
then we also have field_mapping="documents" that makes the  action response’s to map the result to this title as key.
then I also changed display_field="name" from label_field so that for each doc we receive we use the name field as what we see in the way like document.name as we discussed before I statred implementing. 
but we still use value_field as id so it is the one we use later when taking the value of this field to be used in the db when discussing with the agents.

when testing locally i crated an agent that learned a certain doc and one that didn't and asked themt he same question and received the data from the doc only from the related agent so the flow should work well (and of course will be tested more)

i didn't go into the other fields that I didn't change and that are still what existed before but I can if you want - let me know

odaiodeh: @MayaCrmi  something not adding up.
we have field mapping documents, it means the output of the action will go to the document field in config.
now the output is a dict, so we have display and we have value right?
so display you put the name
and for value you put the ids.
so documents should get values, then why it was changed in backed that it gets dicts and not a list of str?
like whats the meaning here for the value if you send the full dict to backend?

MayaCrmi: okay i understand you're asking aboutbthe Union part and not just the naming which is what I fosuced on? if so then the fact that BE receives the option was for like backward compabaility but since i'll run the migration the script i actually think we might not need it. I can change it to like have a class DocInstance with id and name, and make doc_ids receive List[DocInstance] and i think it will work okay. it also will mean we can remove fallbacks in some places acutally. if this is what you meant let me know, i'll test it properly locally and push my changes for your check for it

odaiodeh: no i mean, before the dataflow retriever config was taking a list of str, these lst of strs were ment to be taken from the value part we declared in the Hint, what do you think?

MayaCrmi: i understand what's you're saying about what what value_field means in the hint, that was the original way it worked here. we're using full objects now as the main goal of this PR and the change we decided to make for the sturcture, and it's not a change i made in the meanings of the hints here.
the UI keeps the full object for better user facing presentation of data and just sends that object back, and when needed iit can still uses value_field to get the id.
in general i think it comes down to deciding which contract we want. if we keep it as it now then we expect the ui to only send ids back but that contradicts the goal of this fix, or i can try and make the change explcitly in the backend to be aligned with storing the full object, but i don't now if and how if affects other fields that use this sort of key.

MayaCrmi: ok so display_field is what the user sees and value_field is what gets saved. display_name doesn’t do anything for the purpose of the ticket itself, it's more of just something that annoyed me so i changed the label of the field to display this (documents) instead of doc_ids in the ui modal just becuse we don't longer show the ids at all. 
we decided here to not keep only ids as to make sure the ui always displays names, but the display name isn't related to this change logic from behind the scenes.
if you could maybe just help me by saying exactly what's the problem i can try and change my code or logic to fix it but i'm not sure which part or what is the one you're trying to fully unerstand, display_name or the doc_ids initalization or just understanding things as a whole?

MayaCrmi: ohh i understand the concern then. 
i guess the backend change isn’t  actually required for dataflow itself and it’s purpose is to make sure the config contract is aligned with what the ui now sends. so in the ui, FieldPopulation uses value_field to save the selected ids, and it sends the full selected objects ({id, name}) so the name is persisted and doesn’t depend on any refetching but always saved in the db (as we agreed). if the backend kept  list of str only, we’d lose the name and then need a second call to look up the name every time just to render the ui (if this is how we would've decided to go with this)
so the change is about what we store in the config.. like retrieval logic later still only uses ids since that's how it's been done so far and i didn't want to change it. i don't know but maybe the config can ignore what the ui is expecting but since it's shared between the two i think it should taken into account here as well

multi-agent/elements/retrievers/docs_dataflow/docs_dataflow_retriever.py
```diff
@@ -10,38 +10,16 @@ class DocsDataflowRetriever(BaseRetriever):
...
-        
-        return result if result else None
-
     def __init__(
             self,
             top_k_results: int,
             threshold: float,
             timeout: float = 30.0,
-            doc_ids: Optional[List[Union[dict, str]]] = None,
+            docs: Optional[List[Union[dict, str]]] = None,
```
odaiodeh: same here

PR #46, opened 2026-01-05
Timing Issue | Validation | Workflows selection

Nirsisr:
Timing issue with my code, Under the `Agentic AI Workflows` page, each time user change between different workflows my system triggers 2 API calls.

`
`blueprint.validate` API vall and another API call for `blueprints/available.blueprints.resolved.get?userId={}`.
`

The problem is that sometimes when the chosen blueprint is really thick (in terms of the number of elements it's composed of) the validation API can finish very fast, faster than the other API calls, once all the APIs happening parallely which raise a timing bug.

What happens in that in some scenarios the validation return False, therefore we expect to see the `Load Workflow` btn disabled, and also to see warning btn attached to each Node in the graph that is consider invalid. But due to the timing issue what happens is that the `Load Workflow` btn is ENABLED and we can't see any nodes in the graph with ERRORS (warning icons attached to them) it seems to happen since the @/home/cloud-user/Projects/unifai/ui/client/src/components/agentic-ai/graphs/ReactFlowGraph.tsx doesn't get the response from the validation API, same has happen with the `Load Workflow` btn which doesn't turn out to be DISABLED probably since he doen't get the STATE right.

In other scenarios when all worked as expected, first the `Load Workflow` btn showed with a loader than it turns out to ENABLED / DISABLED, when the timing issue I described happens, note that the btn never turned out to appear as `Loading..` since it also probably doesn't get LOADING state which lead to the entire timing error.

### DIFF
--- a/ui/client/src/components/agentic-ai/graphs/ReactFlowGraph.tsx
+++ b/ui/client/src/components/agentic-ai/graphs/ReactFlowGraph.tsx
@@ -749,6 +749,18 @@ export default function ReactFlowGraph({
     Record<string, NodeStatus>
   >({});
   
+  // Refs to hold latest validation data - allows accessing current values during async operations
+  // This solves the race condition: when graph loading completes, we can immediately apply
+  // whatever validation results are available, without needing an intermediate state
+  const validationResultsRef = useRef(validationResults);
+  const isValidatingRef = useRef(isValidating);
+  
+  // Keep refs synchronized with props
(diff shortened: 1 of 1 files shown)


Nirsisr:
> the code itself looks good, counting on you that it fixes the problem. if you want i can try and run this branch on my local host and see if it fixes the problem we had with that one specific workflow of mine

Sounds like a good plan to me.

PR #49, opened 2026-01-12
Genie 1131/story/support multiple mcp providers

odaiodeh:
**Nir Notes:**

- Supporting selection of 2 providers under the same agent, which will allow the agent to access tools from multiple MCPs servers, same as for tools, the use case raised from `The Source` MCP integration into UnifAI which needed extra capabilities such as GWS Agent that can read spreadsheet which contain inner data from certain `The Source` pages.

- When user create new elements tagged by `VALIDATION` due to APIHint or due to attached Action with Validation assuming the field is not required we don't want to mark it with X, and don't want to block save.  Also, not trigger APIHint behind the scene or the validation action when the component first mounted and those values are wheren't chosen. 

- A2A_Card | Without $ref prefix (bug that appear after last PR)

- Disabled save btn in `edit mode` under our inventory when validation tasks taking place

- Disabled Configuration Page

- Fix bug from prev MR once chat sessions aren't updated with latest data, when there are new Q/A on certain chat (than user moved between different chats and come back again to the first chat) due to cache issue

### DIFF
--- a/ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
+++ b/ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
@@ -94,6 +94,7 @@ export const ElementForm: React.FC<ElementFormProps> = ({
   useEffect(() => {
     if (elementSchema && isOpen) {
       const initialData: any = {};
+      const fieldsToValidate = new Set<string>();
 
       // Set default values from combined schema, excluding hidden fields
       Object.entries(elementSchema.config_schema.properties).forEach(
@@ -149,11 +150,31 @@ export const ElementForm: React.FC<ElementFormProps> = ({
             } else {
               initialData[key] = value;
             }
+
+            // Track fields with validation hints that have values - these will trigger validation
+            // and Save button should be disabled until validation completes
+            const hasValidationHint = fieldSchema?.hints?.action?.hint_type === 'validate' || 
+                                      fieldSchema?.hints?.api?.hint_type === 'validate';
+            if (hasValidationHint) {
+              const processedValue = initialData[key];
+              const hasValue = Array.isArray(processedValue) 
+                ? processedValue.length > 0 
+                : processedValue !== undefined && processedValue !== null && processedValue !== '';
+              if (hasValue) {
+                fieldsToValidate.add(key);
+              }
+            }
           });
         }
       }
 
       setFormData(initialData);
+      
+      // In edit mode, mark fields with validation hints and values as "validating"
+      // This ensures Save button is disabled until all validation API calls complete
+      if (editingElement && fieldsToValidate.size > 0) {
+        setValidatingFields(fieldsToValidate);
+      }
     }
   }, [elementSchema, editingElement, isOpen]);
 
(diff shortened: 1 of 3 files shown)


Nirsisr (approved):

multi-agent/api/flask/endpoints/resources.py
```diff
@@ -157,6 +157,54 @@ def validate_resource(resource_id, timeout_seconds):
...
+    
+    # Validate input
+    if not resource_ids:
+        return jsonify([]), 200
+    
+    # Cap max_workers to prevent resource exhaustion
+    max_workers = min(max_workers, 20)
+    
+    try:
+        results = svc.validate_resources(
```
Lina-AbuYousef: what validation are we doing in this case?

multi-agent/resources/service.py
```diff
@@ -169,6 +170,98 @@ def validate_resource(
...
+            }
+            
+            # Collect results as they complete, placing in correct position
+            for future in as_completed(future_to_index):
+                idx = future_to_index[future]
+                results[idx] = future.result()
+        
+        return results
+
+    def _validate_resource_safe(
```
Lina-AbuYousef: so now instead of direct call and use for the old validate_resource method in case of one single resource, we will use this new method to add the extra protection layer of errors? no direct call to validate_resource ? only by _validate_resource_safe

ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
```diff
@@ -140,11 +150,31 @@ export const ElementForm: React.FC<ElementFormProps> = ({
             } else {
               initialData[key] = value;
             }
+
+            // Track fields with validation hints that have values - these will trigger validation
+            // and Save button should be disabled until validation completes
+            const hasValidationHint = fieldSchema?.hints?.action?.hint_type === 'validate' ||
```
Lina-AbuYousef: I think it’s time to schedule a meeting to discuss introducing a smart analysis function.
This function would take a fieldSchema as input and return all related data—such as hint types and any other derived information—by building a single, well-defined object that fully describes the field.

The idea is to use this object consistently across the codebase, instead of performing repeated conditional checks in multiple places.

This would also make the UI more resilient to future changes. If the schema structure returned from the multi-agent backend changes, we would only need to update the logic in one place, rather than tracking and fixing many scattered usages.

For example, we could expose a simplified object like:
hintType: "validate", etc.

multi-agent/resources/service.py
odaiodeh: validate resource call _validate_resource_safe now

multi-agent/api/flask/endpoints/resources.py
Nirsisr: Lina each call to `validaiton` whether it's per resource or with multiple will update the cache.

Now since we have multiple providers per agent, we need to have validation for multi-resources in parallel.

multi-agent/elements/llms/google_genai/schema_sanitizer.py
```diff
@@ -0,0 +1,95 @@
...
+        These typically represent internal service dependencies from MCP
+        servers that shouldn't be exposed to the LLM.
+        """
+        if prop_value is None:
+            return True
+        if not isinstance(prop_value, dict):
+            return False
+        if not prop_value:
+            return True
+        if prop_value.keys() == {'title'}:
```
MayaCrmi: is it possible for the prop to have more key "title" and be invalid? trying to understand how and why this type of dict can be created

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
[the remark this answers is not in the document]
Nirsisr: Yep Maya's, it solved both here, and under @ElementForm (code changes in both areas solved the issue)

PR #115, opened 2026-03-09
Add clone workflow feature

ItzikEzra-rh:
## Summary
  - Adds a Clone button next to Share/Edit/Delete in the workflows list
  - Reuses the existing ShareCloner.clone_blueprint() with sender==recipient
  - Cloned workflow gets '(copy)' suffix, all resources and step UIDs are regenerated

  ## Changes
  - **multi-agent/api/flask/endpoints/blueprints.py** - New POST /blueprint.clone endpoint
  - **multi-agent/sharing/cloner.py** - Smart naming: '(copy)' for self-clone, '(from
  user)' for sharing
  - **ui/client/src/api/blueprints.ts** - New cloneBlueprint() API function
  - **ui/client/src/components/agentic-ai/WorkflowsPanel.tsx** - Clone button + handler

  ## How it works
  The existing ShareCloner already handles blueprint cloning with full dependency
  resolution, RID remapping, and step UID regeneration. For 'clone to self', we simply
  call it with sender_user_id == recipient_user_id.

  4 files changed, ~92 insertions, ~5 deletions.

## Release Notes

* **New Features**
  * Added blueprint cloning capability, enabling users to quickly duplicate workflows along with all associated dependencies and configurations.
  * Introduced Clone button in the workflow interface with comprehensive state management and error handling for reliable cloning operations.
  * Built-in operation tracking prevents accidental duplicate clone submissions during concurrent operations.

### DIFF
--- a/ui/client/src/api/blueprints.ts
+++ b/ui/client/src/api/blueprints.ts
@@ -170,6 +170,25 @@ export async function updateBlueprint(
 }
 
 // ────────────────────────────────────────────────────────────────────────────────
+// Blueprint Cloning
+// ────────────────────────────────────────────────────────────────────────────────
+
+/**
+ * Clone a blueprint and all its dependencies for the current user.
+ * Creates a new blueprint with "(copy)" suffix.
+ */
+export async function cloneBlueprint(
+  blueprintId: string,
+  userId: string
+): Promise<SaveBlueprintResponse> {
+  const { data } = await axios.post<SaveBlueprintResponse>('/blueprints/blueprint.clone', {
+    blueprintId,
+    userId,
+  });
+  return data;
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
 // Blueprint Metadata & Sharing
 // ────────────────────────────────────────────────────────────────────────────────
 
--- a/ui/client/src/api/blueprints.ts
+++ b/ui/client/src/api/blueprints.ts
@@ -170,18 +170,18 @@ export async function updateBlueprint(
 }
 
 // ────────────────────────────────────────────────────────────────────────────────
-// Blueprint Cloning
+// Blueprint Duplication
 // ────────────────────────────────────────────────────────────────────────────────
 
 /**
- * Clone a blueprint and all its dependencies for the current user.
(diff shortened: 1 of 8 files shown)


Nirsisr (dismissed):
Itzik great work with leveraging our current infra for the implementation, kudos.

Nirsisr (dismissed):

odaiodeh (dismissed):

ui/client/src/api/blueprints.ts
```diff
@@ -169,6 +169,25 @@ export async function updateBlueprint(
...
+
+/**
+ * Clone a blueprint and all its dependencies for the current user.
+ * Creates a new blueprint with "(copy)" suffix.
+ */
+export async function cloneBlueprint(
+  blueprintId: string,
+  userId: string
+): Promise<SaveBlueprintResponse> {
+  const { data } = await axios.post<SaveBlueprintResponse>('/blueprints/blueprint.clone', {
```
MayaCrmi: i'm not sure if we need a new endpoint for this or if it's better to use one endpoint for all cloning logic with either shared or self copy. what do you think?

MayaCrmi (commented):
great job with this.
the only thing i think maybe the better way to call this UI-wise is 'duplicate' instead of 'clone'? i think it's a more instinctive naming for it for users

ui/client/src/api/blueprints.ts
ItzikEzra-rh: Good point. I think a separate endpoint still makes sense, share is async/two-step (invite → accept/decline) while duplicate is instant and self-service.
Merging them would either force self-clone through the invite flow (creating unnecessary invite records) or break the consent model for sharing.

But you're right that the current implementation is inconsistent, the clone endpoint bypasses ShareService and calls the cloner directly.
I'll add a duplicate_blueprint() method on ShareService so both flows go through the same service layer.

MayaCrmi: that's great, I agree. Thanks!

MayaCrmi (approved):
nice :) I think we talked about moving the duplicate to the shares.ts logic but I think I agree with it being under blueprints.

PR #123, opened 2026-03-17
GENIE-827/story/implement-logout

MayaCrmi:
The logout button in the header didn't do anything. When I tried to make the button use the logout method we already had, it cleared the app session but Keycloak's SSO session just kept re-authenticating the user, making the logout impossible. To solve this I had to change some logic under the authentication files like this:

auth_manager.py: Logout revokes the Keycloak session server-side, and login endpoint accepts an optional prompt=login parameter to force Keycloak to show a login form instead of just re-authenticating
AuthContext.tsx: After logout, a unifai_logged_out flag is set so the app knows to show a logged-out screen instead of auto-redirecting to login. When the user clicks "Log in" from that screen, prompt=login is passed to force a login form.
authClient.ts: The 401 now skips the auto-redirect logic when the user is in a logged-out state.

Other changed files are more straight-forward.

image

(not 100% sure about phrasing here so let's discuss!)
after signing out:

image

## Summary by CodeRabbit

* **New Features**
  * Added a full-screen Login page and a dedicated "Signed out" screen with a login action.

* **Bug Fixes**
  * Prevents automatic auth redirects for already-logged-out sessions.
  * Sign-out state now syncs across browser tabs.
  * Unauthenticated flows show the login UI instead of forcing an immediate redirect.

* **Style**
  * Updated loading spinner and message colors; added smooth color transitions to logout/login buttons.


sfiresht (commented):
looks great and one of the much needed features.

yhabushi79 (dismissed):
you already fix the issue

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -139,8 +145,25 @@ def auth_callback():
...
+            if refresh_token_val:
+                try:
+                    keycloak_base_url = config.keycloak_base_url
+                    realm = config.get('keycloak_realm', 'master')
+                    logout_url = f"{keycloak_base_url}/realms/{realm}/protocol/openid-connect/logout"
+                    http_requests.post(logout_url, data={
+                        'client_id': config.client_id,
+                        'client_secret': config.client_secret,
+                        'refresh_token': refresh_token_val,
+                    }, timeout=10)
```
sfiresht, line 162: is there a reason to use this as a simple http request? isn't there a way to use the keycloak module and logout?

MayaCrmi:
will re open once the new hardcoded users logic is merged here.

PR #130, opened 2026-03-23
Genie 1014/story/rputer default part orch

Lina-AbuYousef:
### Summary

Previously, to build a workflow with an orchestrator, users had to manually find the `router_direct` condition in the sidebar's "Conditions" tab, drag it onto the orchestrator node, and then draw branch edges. This PR makes the `router_direct` condition an automatic, system-managed part of the orchestrator — when a user draws an edge from an orchestrator node, it's automatically treated as a branch edge with direct routing. No manual condition attachment required.

### How the Solution Works

The orchestrator's config now carries a `router_direct: true` flag (hidden from the UI form). On the **backend**, the config schema was updated to accept this flag, the `router_direct` condition spec was fixed so it registers correctly in the `ElementRegistry`, and validation was relaxed so orchestrators without edges yet don't fail. On the **frontend**, the UI detects the `router_direct` flag on the node's config when drawing edges and automatically creates branch edges, injecting an inline `router_direct` condition definition into the YAML draft — producing the same YAML structure as existing fixtures. No auto-injection logic was added to the backend resolver or plan builder; the UI generates the full blueprint directly.

#### Backend (`multi-agent`)

| File | Change |
|------|--------|
| `elements/nodes/orchestrator/config.py` | Added `router_direct: bool = True` field to `OrchestratorNodeConfig` with `HiddenHint` so it's accepted by Pydantic but hidden from the UI form. This is the signal the UI reads to determine routing behavior. |
| `elements/conditions/router_direct/spec/spec.py` | Added missing `HiddenHint` import and `hints = [HiddenHint(...)]` to mark the condition as hidden in the catalog. Without the import, the spec module failed silently during auto-discovery and the condition was never registered in the `ElementRegistry`. |
| `graph/validation/orchestrator/checker.py` | Relaxed Rule 3 (missing finalize check): orchestrators with no branches yet now skip validation instead of erroring. Since the router is auto-attached, the orchestrator may exist on the canvas before any edges are drawn. |
| `graph/rt_graph_plan.py` | Deep-copy node and condition instances when creating runtime steps, so multiple steps sharing the same resource ID get independent objects (prevents `set_context` overwrites). |

#### Frontend (`ui/client`)

| File | Change |
|------|--------|
| `hooks/use-graph-logic.ts` | **`onConnect`**: Added `isRouterDirect` check — if the source node's config has `router_direct === true`, the branch modal opens with `conditionType: "router_direct"` (no explicit condition needed in `referencedConditions`). **`createConditionalEdge`**: For orchestrator nodes, sets `exit_condition: "router_direct"` on the step and adds a shared inline `router_direct` condition definition to the YAML conditions array (deduplicated). For non-orchestrator nodes with explicit conditions, behavior is unchanged. |
| `workspace/BuildingBlocksSidebar.tsx` | Removed the "Conditions" tab entirely. The sidebar now only shows "Nodes". Removed `conditions` prop, `Tabs`/`TabsContent` components, condition rendering, condition-related theme colors, and drag-drop instructions for conditions. |
| `workspace/NewGraph.tsx` | Removed `conditionsData` and `allBlocksData` from the graph hook. `usedElementIds` no longer tracks attached conditions. `BuildingBlocksSidebar` no longer receives the `conditions` prop. |

### DIFF
--- a/ui/client/src/workspace/BuildingBlocksSidebar.tsx
+++ b/ui/client/src/workspace/BuildingBlocksSidebar.tsx
@@ -1,27 +1,22 @@
 import React, { useState, useMemo } from "react";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
-import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { Eye, Loader2 } from "lucide-react";
 import { BuildingBlock } from "@/types/graph";
 import { getCategoryDisplay } from "@/components/shared/helpers";
 import ResourceDetailsModal from "./ResourceDetailsModal";
-import { UmamiTrack } from '@/components/ui/umamitrack';
-import { UmamiEvents } from '@/config/umamiEvents';
 import { useTheme } from "@/contexts/ThemeContext";
 import { deriveThemeColors } from "@/lib/colorUtils";
 
 interface BuildingBlocksSidebarProps {
   buildingBlocks: BuildingBlock[];
-  conditions: BuildingBlock[];
   isLoading: boolean;
   onDragStart: (event: React.DragEvent, block: BuildingBlock) => void;
-  usedElementIds?: Set<string>; // Track which elements are currently used on canvas
+  usedElementIds?: Set<string>;
 }
 
 const BuildingBlocksSidebar: React.FC<BuildingBlocksSidebarProps> = ({
   buildingBlocks,
-  conditions,
   isLoading,
   onDragStart,
   usedElementIds = new Set<string>(),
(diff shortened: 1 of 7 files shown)


MayaCrmi (commented):
my main question is do we agree to not let other agents to use routing logic at all anymore?

ui/client/src/workspace/BuildingBlocksSidebar.tsx
```diff
@@ -1,57 +1,36 @@
 import React, { useState, useMemo } from "react";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
-import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { Eye, Loader2 } from "lucide-react";
 import { BuildingBlock } from "@/types/graph";
 import { getCategoryDisplay } from "@/components/shared/helpers";
 import ResourceDetailsModal from "./ResourceDetailsModal";
-import { UmamiTrack } from '@/components/ui/umamitrack';
```
sfiresht, line 9: I see that the tracking was removed, isn't there a way for the user to add conditions to a  node?

sfiresht:
@Lina-AbuYousef  will you come back to this feature?  if not maybe it's better to close this PR.

PR #133, opened 2026-03-25
Genie 1448/bugfix/add from sender on elements

MayaCrmi:
Self explanatory :) This resolves following ticket:
"  Shared workflow gets "(from sender)" suffix but inventory objects (agents, conditions, orchestrators) do not — inconsistent naming  causes confusion"
i also noticed that on my laptop screen the Agentic Chats title is a bit hidden so added some css to improve that.

This needs to be tested on PP properly still.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed layout and scrolling behavior in execution panel sidebars and modal windows, ensuring proper content sizing and full functionality of interactive elements.

* **New Features**
  * Enhanced resource naming in sharing operations to provide clearer identification of cloned resources and their origin.


Lina-AbuYousef (commented):
looks good, just small question, is it better to have the sender seperated and not with cache_data?

Nirsisr (dismissed):

Lina-AbuYousef (dismissed):

odaiodeh (changes requested):
Check the comment on the design layers and responsibility of each layer

multi-agent/lib/mas/sharing/service.py
```diff
@@ -215,3 +220,24 @@ def _can_delete_invite(self, invite: ShareInvite, user_id: str) -> bool:
...
+        if "@" in value:
+            value = value.split("@", 1)[0]
+
+        if re.fullmatch(r"[0-9a-fA-F]{24,}", value):
+            value = value[:8]
+
+        if len(value) > MAX_DISPLAY_LEN:
+            value = value[:MAX_DISPLAY_LEN]
+
+        return value or "shared"
```
odaiodeh: @MayaCrmi could you please explain what this function does and why it’s needed?

Also, I’m trying to understand the use of @ coming from the UI. In what scenarios would we receive @ from the UI, and why would that be considered valid behavior?

If @ is coming from the UI, does that mean it’s derived from the user_id? If so, in which cases would that make sense?

odaiodeh (changes requested):

multi-agent/lib/mas/sharing/service.py
MayaCrmi: it was a coderabbit suggestion of making sure the usernames are clean. nir and i talked about it a bit. we won't need this as long as all of our users are from sso, we can remove it if you want, it's more a future backup vibe

odaiodeh: @MayaCrmi yes please, lets remove it. the display name can be derived from the userid + "from", and no need for it to be in the context. maybe you create display name function that adds to the user id the "from"

odaiodeh (approved):

Lina-AbuYousef (approved):

Nirsisr (approved):

PR #134, opened 2026-03-26
Pre-defined Templates & Admin Template CRUD

Lina-AbuYousef:
### Overview

Adds a template management system with pre-defined workflow templates that are seeded on startup, plus admin-only CRUD operations for creating and deleting templates via the UI and API.

### What's New

**Template Seeding & Fixtures**
- YAML-based fixture templates are automatically seeded into MongoDB on application startup
- Seeder respects soft-deleted templates (won't re-insert templates an admin has removed)
- Per-file error handling ensures one bad fixture doesn't block the rest

**Soft-Delete for Templates**
- Template deletion is a soft-delete (`deleted: True`)
- All read operations (`get`, `list`, `search`, `count`) exclude soft-deleted templates
- `exists()` supports `include_deleted` parameter for seeder compatibility

**Admin CRUD (Backend)**
- `POST /templates/template.create` — Create a new template (admin only)
- `DELETE /templates/template.delete` — Soft-delete a template (admin only)
- Both endpoints are protected with `@require_admin_access`

**Admin UI**
- Admin users see "Add Template" button in the template catalog
- Create Template dialog with tabs for Metadata, Blueprint (YAML), and Placeholders (JSON)
- View YAML dialog for inspecting a template's blueprint definition
- Delete template with confirmation dialog
- Admin status is checked via the existing `useAdminAccess()` hook

---

#### Template 1: "The L1/L2 SRE Auto-Medic"

| | |
|---|---|
| **Agents Used** | OpenShift SRE, RHEL SRE, Jira, Reporter |
| **Value Proposition** | Zero-touch incident triage |
| **Scenario** | An OpenShift pod crashes. The orchestrator routes the logs to the OpenShift agent, asks the RHEL agent to check host-level CPU/Memory, creates a Jira ticket with a root-cause hypothesis, and produces a structured incident report. |
| **Manual Execution** | 45–60 minutes — A human SRE context-switches, logs into multiple dashboards, pulls logs, correlates timestamps, and writes up a Jira ticket. |
| **Autonomous Execution** | ~90 seconds — The human starts their investigation with the root-cause hypothesis already in hand. |

#### Template 2: "The Autonomous Product Owner"

| | |
|---|---|
| **Agents Used** | Confluence, Jira, Git |
| **Value Proposition** | From PRD to Sprint Backlog in 60 seconds |
| **Scenario** | Point the Orchestrator at a PRD in Confluence. It extracts the requirements, asks the Git agent to check the current repo structure, and automatically breaks the PRD down into perfectly formatted Jira Epics and Stories. |
| **Manual Execution** | 4–6 hours — A Product Manager or Lead Engineer manually reads the document, breaks down the technical tasks, and creates dozens of individual Jira tickets. |
| **Autonomous Execution** | ~3 minutes — The team simply reviews and approves the generated backlog. |

#### Template 3: "The Continuous Compliance Auditor"

| | |
|---|---|
| **Agents Used** | Git, Google Workspace, Reporter |
| **Value Proposition** | Never fail a SOC2 audit again |
| **Scenario** | Automatically scans PRs in Git, cross-references them against security policies in Google Docs, and generates a weekly compliance report. |
| **Manual Execution** | 8–12 hours per audit cycle — Engineering managers manually sample PRs, check approval histories, and cross-reference them against static policy docs. |
| **Autonomous Execution** | Continuous tracking + ~2 minutes for report generation. 100% coverage instead of manual sampling, with zero human toil. |

---

### Files Changed

| Area | Files | Summary |
|---|---|---|
| **Repository Soft-Delete** | `template_repository.py`, `repository.py` | Soft-delete semantics, `exists(include_deleted)`, active-only filtering |
| **Fixture Seeding** | `seeder.py`, `container.py` | YAML fixture loader with shape validation, integrated into `AppContainer` startup |
| **Fixture Templates** | `run/fixtures/templates/*.yml` | Three pre-defined blueprint YAML fixtures |
| **Backend API** | `templates.py` | `template.create` and `template.delete` endpoints with `@require_admin_access` |
| **Frontend API** | `templates.ts` | `createTemplate` and `deleteTemplate` client methods |
| **UI Dialogs** | `CreateTemplateDialog.tsx`, `ViewYamlDialog.tsx` | Create template form and read-only YAML viewer |
| **UI Admin Controls** | `TemplateCatalog.tsx`, `TemplateCard.tsx` | Admin-only buttons gated by `isAdmin` |
| **Page Integration** | `AgenticTemplates.tsx` | Wired dialogs, API calls, confirmation flows, and toasts |


multi-agent/bootstrap/container.py
```diff
@@ -174,8 +175,19 @@ def __init__(self, cfg: AppConfig):
...
         )
 
+        self._seed_templates()
+
         self._initialized = True
 
+    def _seed_templates(self):
+        """Load YAML template fixtures into the database if not already present."""
+        import logging
+        logger = logging.getLogger(__name__)
```
MayaCrmi: is it better to import it here or should it be moved to the top of the file? like why only do it in this function?

multi-agent/run/fixtures/seeder.py
```diff
@@ -0,0 +1,79 @@
...
+    TemplateMetadata,
+)
+from mas.templates.repository.repository import TemplateRepository
+
+logger = logging.getLogger(__name__)
+
+FIXTURES_DIR = Path(__file__).parent / "templates"
+
+
+def seed_templates(repository: TemplateRepository) -> int:
```
MayaCrmi, line 29: the insertion of templates is by upload of a yaml file? do we display a template for how it should look to the users?
also in the description it says only admins can create temps, where do we check if the user's an admin?

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  const handleAddTag = () => {
+    const trimmed = tagInput.trim();
+    if (trimmed && !tags.includes(trimmed)) {
+      setTags([...tags, trimmed]);
+      setTagInput('');
+    }
+  };
+
+  const handleAddCapability = () => {
```
MayaCrmi: don't have to i guess but is there a way to combine this and the AddTag function since they're the same functionality?

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+      version,
+      tags,
+      output_capabilities: capabilities,
+      is_public: true,
+    };
+
+    await onSubmit({ draft, placeholders, metadata });
+  };
+
+  const handleClose = () => {
```
MayaCrmi, line 173: in the onAddTag etc functions you do setTag(''), i think you should also do it here

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    <Badge key={tag} variant="secondary" className="text-xs bg-primary/10 text-primary">
+                      #{tag}
+                      <button onClick={() => setTags(tags.filter(t => t !== tag))} className="ml-1 hover:text-white">
+                        <X className="h-2.5 w-2.5" />
+                      </button>
+                    </Badge>
+                  ))}
+                </div>
+              )}
+            </div>
```
MayaCrmi, line 259: can you maybe add a screenshot of how it looks now?

ui/client/src/components/agentic-ai/templates/TemplateCard.tsx
```diff
@@ -43,6 +49,35 @@ export const TemplateCard: React.FC<TemplateCardProps> = ({
                 </Badge>
               </div>
             </div>
+
+            {isAdmin && (
```
MayaCrmi, line 53: if i underdstand correctly you check here if the user's an admin and then the rest of the funcionality isn't limited, i think we should still cover our bases with the require_admin decorator we have on the apis

ui/client/src/pages/AgenticTemplates.tsx
```diff
@@ -62,15 +78,13 @@ export default function AgenticTemplates() {
   const handleGenerateWorkflow = useCallback(async (data: TemplateFormData) => {
     if (!selectedTemplate || !user) return;
     
-    // Generate a blueprint name based on template name
     const blueprintName = `${selectedTemplate.name} - ${new Date().toLocaleDateString()}`;
```
MayaCrmi, line 80: i think we should bring back the comments that the cursor removed from this file

multi-agent/lib/mas/templates/repository/repository.py
```diff
@@ -40,9 +40,11 @@ def update(self, template: Template) -> bool:
     @abstractmethod
     def delete(self, template_id: str) -> bool:
         """
-        Delete a template by ID.
+        Soft-delete a template by ID (marks deleted=True).
```
sfiresht, line 43: is there any retention? if not it means that every template ever created lasts forever in the DB, right? 
what about templates that were remomved (not necessarily through the UI but in the dev process, or changed for that matter? do we have tempate versioning in place? (I know there's a talk about it here and there but just want to make sure we're aware ..)

multi-agent/run/fixtures/seeder.py
```diff
@@ -0,0 +1,79 @@
...
+            with open(path, "r") as f:
+                raw = yaml.safe_load(f)
+
+            template_id = raw.get("template_id")
+            if not template_id:
+                logger.warning("Skipping %s — missing 'template_id' at top level.", path.name)
+                continue
+
+            if repository.exists(template_id):
+                logger.debug("Template '%s' already exists — skipping.", template_id)
```
sfiresht: generally, I think it's healthier to have messages starting at .info in the log (though the number of messages here will be very low, but still).

ui/client/src/api/templates.ts
```diff
@@ -51,6 +51,41 @@ export async function getTemplateSchema(templateId: string): Promise<TemplateInp
...
+/**
+ * Delete a template by ID (admin only)
+ * DELETE /templates/template.delete
+ */
+export async function deleteTemplate(templateId: string): Promise<DeleteTemplateResponse> {
+  const response = await axios.delete<DeleteTemplateResponse>('/templates/template.delete', {
+    params: { templateId }
+  });
+  return response.data;
+}
```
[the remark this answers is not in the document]
Lina-AbuYousef: right, adding the require_admin_access decorator to the endpoints

multi-agent/run/fixtures/seeder.py
```diff
@@ -0,0 +1,79 @@
...
+                created_at=now,
+                updated_at=now,
+            )
+
+            repository.save(template)
+            inserted += 1
+            logger.info("Seeded template '%s' from %s", template_id, path.name)
+
+        except Exception:
+            logger.exception("Failed to seed template from %s", path.name)
```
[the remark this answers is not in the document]
Lina-AbuYousef: very nice bug catch, keep catching these bugs, i'm fixing

multi-agent/bootstrap/container.py
```diff
@@ -174,8 +175,19 @@ def __init__(self, cfg: AppConfig):
...
+        try:
+            seed_templates(self.template_repo)
+        except Exception:
+            logger.exception("Template seeding failed — continuing startup.")
```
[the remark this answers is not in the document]
Lina-AbuYousef: that's over-engineering for fixture seeding.

multi-agent/adapters/outbound/mongo/template_repository.py
```diff
@@ -68,20 +68,23 @@ def update(self, template: Template) -> bool:
...
+        """Load a non-deleted template by ID."""
+        doc = self._col.find_one({"template_id": template_id, "deleted": {"$ne": True}})
         if not doc:
             raise KeyError(f"Template not found: {template_id}")
         return self._doc_to_template(doc)
 
     def exists(self, template_id: str) -> bool:
-        """Check if a template exists."""
+        """Check if a template exists (including soft-deleted — used by seeder)."""
         return self._col.count_documents({"template_id": template_id}, limit=1) == 1
```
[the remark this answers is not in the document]
Lina-AbuYousef: Already fixed. This was the core of the exists() change we just made.

multi-agent/adapters/outbound/mongo/template_repository.py
```diff
@@ -68,20 +68,23 @@ def update(self, template: Template) -> bool:
...
     def delete(self, template_id: str) -> bool:
-        """Delete a template by ID."""
-        res = self._col.delete_one({"template_id": template_id})
-        return res.deleted_count == 1
+        """Soft-delete a template (set deleted=True so seeder won't re-insert)."""
+        res = self._col.update_one(
+            {"template_id": template_id},
+            {"$set": {"deleted": True, "updated_at": datetime.now(timezone.utc)}},
+        )
+        return res.modified_count == 1
```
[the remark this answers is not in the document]
Lina-AbuYousef: Also already fixed. The delete() selector now includes the soft-delete guard:

multi-agent/run/fixtures/seeder.py
```diff
@@ -0,0 +1,79 @@
...
+
+    inserted = 0
+    for path in yaml_files:
+        try:
+            with open(path, "r") as f:
+                raw = yaml.safe_load(f)
+
+            template_id = raw.get("template_id")
+            if not template_id:
+                logger.warning("Skipping %s — missing 'template_id' at top level.", path.name)
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+      return;
+    }
+
+    let placeholders: Record<string, any>;
+    try {
+      placeholders = JSON.parse(placeholdersJson);
+    } catch (e: any) {
+      setParseError(`Placeholders JSON parse error: ${e.message}`);
+      return;
+    }
```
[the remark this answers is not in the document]
Lina-AbuYousef: Done. Now arrays, primitives, and null are rejected with a clear error message before reaching onSubmit.

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+      setDraftYaml('');
+      setPlaceholdersJson('{\n  "categories": []\n}');
+      setAuthor('');
+      setCategory('');
+      setVersion('1.0.0');
+      setTags([]);
+      setCapabilities([]);
+      setParseError(null);
+      onOpenChange(false);
+    }
```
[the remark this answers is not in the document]
Lina-AbuYousef: Valid point -- tagInput and capabilityInput are missing from the reset, and if the parent closes via open={false} after a successful submit, handleClose never runs at all.

check the fix and resolve if fixed

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
```diff
@@ -0,0 +1,259 @@
...
+                </Button>
+              </div>
+              {tags.length > 0 && (
+                <div className="flex flex-wrap gap-1 mt-1">
+                  {tags.map((tag) => (
+                    <Badge key={tag} variant="secondary" className="text-xs bg-primary/10 text-primary">
+                      #{tag}
+                      <button onClick={() => setTags(tags.filter(t => t !== tag))} className="ml-1 hover:text-white">
+                        <X className="h-2.5 w-2.5" />
+                      </button>
```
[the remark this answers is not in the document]
Lina-AbuYousef: nice ux improvement , check the fix

ui/client/src/components/agentic-ai/templates/TemplateCard.tsx
```diff
@@ -43,6 +49,35 @@ export const TemplateCard: React.FC<TemplateCardProps> = ({
...
+                  className="h-8 w-8 p-0 text-gray-400 hover:text-red-400"
+                  title="Delete template"
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    onDelete?.(template);
+                  }}
+                >
+                  <Trash2 className="h-4 w-4" />
+                </Button>
+              </div>
```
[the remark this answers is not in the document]
Lina-AbuYousef: Already done -- I just added aria-label="View YAML" and aria-label="Delete template" to both buttons in TemplateCard.tsx alongside the existing title attributes in the previous change.

ui/client/src/components/agentic-ai/templates/ViewYamlDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [draft]);
+
+  const handleCopy = async () => {
+    try {
+      await navigator.clipboard.writeText(yamlContent);
+      setCopied(true);
+      toast({ title: 'Copied', description: 'YAML copied to clipboard.' });
+      setTimeout(() => setCopied(false), 2000);
+    } catch {
```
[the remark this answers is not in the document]
Lina-AbuYousef: Done. The timeout is now tracked in a ref -- previous timers are cleared before creating a new one on rapid clicks, and the useEffect cleanup clears any pending timer on unmount.

ui/client/src/pages/AgenticTemplates.tsx
```diff
@@ -111,6 +124,69 @@ export default function AgenticTemplates() {
...
+      setShowYamlDialog(true);
+    } catch (err) {
+      console.error('Error fetching template for YAML view:', err);
+      toast({
+        title: 'Error',
+        description: 'Failed to load template YAML.',
+        variant: 'destructive'
+      });
+    }
+  }, [toast]);
```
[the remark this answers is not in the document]
Lina-AbuYousef: This is a very edge-case scenario -- admin rapidly clicking "View YAML" on two different templates before the first fetch completes. In practice, the fetch is fast (single document lookup by ID) and there's usually a loading state that prevents interaction. I'd skip this one -- it's low-risk and the fix (adding a ref to track the latest request ID or an AbortController) adds complexity for a scenario that's unlikely to happen in real usage.

multi-agent/run/fixtures/templates/blueprint_autonomous_product_owner.yml
```diff
@@ -0,0 +1,553 @@
...
+          ### Role & Directive
+          You are the "Sprint Backlog Reporting Agent," an elite product management analyst. Your sole purpose is to process the outputs from the PRD extraction, repo analysis, and Jira ticket creation, then produce a polished, executive-ready sprint backlog report.
+
+          ### Phase 1: Deep Reasoning (The `<thinking>` Block)
+          Before writing any part of the report, you MUST enclose your analytical process within `<thinking>` and `</thinking>` XML tags. Inside these tags:
+          1. **Data Ingestion:** Identify all Epics and Stories created by the Jira agent.
+          2. **Coverage Analysis:** Cross-reference created tickets against the PRD requirements — are any missing?
+          3. **Dependency Mapping:** Identify Stories that depend on each other or share modules.
+          4. **Sprint Planning:** Determine a logical sprint ordering based on priorities and dependencies.
+
```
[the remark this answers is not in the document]
Lina-AbuYousef: Done. Updated all three templates:

blueprint_autonomous_product_owner.yml
blueprint_sre_auto_medic.yml
blueprint_continuous_compliance_auditor.yml
In each one: renamed "Phase 1: Deep Reasoning" to "Phase 1: Internal Verification", removed all references to <thinking> tags, and added "Do not include your internal reasoning in the output." The verification steps themselves are preserved -- the LLM still performs them, it just doesn't emit them to the user.

multi-agent/run/fixtures/templates/blueprint_sre_auto_medic.yml
```diff
@@ -0,0 +1,504 @@
...
+  # Providers
+  # ----------------------------
+  providers:
+    - rid: jira_mcp_rid
+      name: Jira MCP PAT
+      type: mcp_server
+      config:
+        type: mcp_server
+        bearer_token: PLACEHOLDER_JIRA_PAT
+        mcp_url: https://jira-mcp-tag-ai--runtime-int.apps.stc-ai-e1-pp.imap.p1.openshiftapps.com/mcp
```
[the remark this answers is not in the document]
Lina-AbuYousef: Done. The SRE template now:

Uses PLACEHOLDER_JIRA_MCP_URL instead of the hard-coded runtime-int URL
Exposes mcp_url as a required placeholder field so users can configure it per deployment
Fixes the stale "base64-encoded" hint on the PAT token

multi-agent/run/fixtures/templates/blueprint_autonomous_product_owner.yml
```diff
@@ -0,0 +1,553 @@
...
+        - rid: jira_mcp_rid
+          placeholders:
+            - field_path: mcp_url
+              required: true
+              label: Jira MCP Server URL
+              hint: "URL of your Jira MCP server endpoint (e.g. https://mcp-atlassian.example.com/mcp)"
+            - field_path: bearer_token
+              required: true
+              label: Jira PAT Token
+              hint: "Your Jira Personal Access Token (base64-encoded username:token)"
```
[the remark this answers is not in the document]
Lina-AbuYousef: Done

multi-agent/run/fixtures/templates/blueprint_sre_auto_medic.yml
```diff
@@ -0,0 +1,504 @@
...
+
+          ## Investigation Playbook
+          When given an incident (pod name, namespace, or symptom), execute these steps systematically:
+
+          1. **Pod Status**: Run `oc get pod <pod-name> -n <namespace> -o wide` to see current state, restarts, node assignment.
+          2. **Pod Events**: Run `oc describe pod <pod-name> -n <namespace>` — look for Events section (OOMKilled, ImagePullBackOff, FailedScheduling, Evicted).
+          3. **Container Logs**: Run `oc logs <pod-name> -n <namespace> --tail=100` for current container. If the pod restarted, also check `oc logs <pod-name> -n <namespace> --previous` for the crashed container's last logs.
+          4. **Resource Limits**: Check if the pod has CPU/memory requests and limits. Look for resource pressure.
+          5. **Namespace Events**: Run `oc get events -n <namespace> --sort-by=.lastTimestamp` to see broader namespace activity around the time of failure.
+          6. **Node Info**: Run `oc describe node <node-name>` if the pod's node is known — check Conditions (MemoryPressure, DiskPressure, PIDPressure) and Allocated resources.
```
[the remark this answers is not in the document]
Lina-AbuYousef: nice catch for the first one!
but for Input validation in OcExecTool -- I'd skip this one. The tool already uses shlex.split() + structured oc.invoke() which prevents injection. Adding regex validation on pod names is defense-in-depth for a hypothetical future change, but it adds maintenance burden and could break legitimate edge cases (some pod names contain underscores, generated hashes, etc.). If the implementation ever changes to use shell execution, that's the appropriate time to add validation.

multi-agent/run/fixtures/templates/blueprint_sre_auto_medic.yml
```diff
@@ -0,0 +1,504 @@
...
+          - **Pod Status**: Current state, restart count, how long it's been running/failing
+          - **Root Cause Indicators**: What the events/logs suggest (OOMKilled, CrashLoopBackOff, etc.)
+          - **Timeline**: When did the issue start? Any correlated events?
+          - **Resource Analysis**: Are limits set? Are they being exceeded?
+          - **Node Assignment**: Which node is the pod on? Any node-level conditions?
+
+          ## Rules
+          - Be thorough: always check logs AND events, not just one.
+          - If the pod name or namespace is ambiguous, search with `oc get pods --all-namespaces | grep <keyword>`.
+          - Report raw evidence — include relevant log snippets and event messages. Don't just summarize.
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

multi-agent/run/fixtures/templates/blueprint_autonomous_product_owner.yml
```diff
@@ -0,0 +1,553 @@
...
+        display_name: Backlog Summary Reporter
+        description: |
+          Produces the final sprint backlog summary report for team review.
+        tags: []
+
+    - uid: finalize
+      node: final_answer_node_rid
+      after: null
+      exit_condition: null
+      branches: null
```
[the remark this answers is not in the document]
Lina-AbuYousef: **Since this pattern works in production across all templates and we don't have test coverage to verify a change won't break the orchestration flow, the safer approach is to leave it as-is. If you want to investigate further, the key code is in base_builder.py's compile_from_plan() and the OrchestratorNode class -- specifically how it handles the work plan and task responses.**

multi-agent/run/fixtures/templates/blueprint_sre_auto_medic.yml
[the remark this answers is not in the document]
Lina-AbuYousef: resolve

multi-agent/bootstrap/container.py
Lina-AbuYousef: Good catch — logging is already imported at the top of the file. Removed the redundant local import.

ui/client/src/components/agentic-ai/templates/TemplateCard.tsx
Lina-AbuYousef: Added @require_admin_access decorator to both template.create and template.delete Flask endpoints, matching the pattern used in statistics.py.

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
Lina-AbuYousef: Fixed — extracted a resetForm() function that clears all fields including tagInput and capabilityInput. Also added a useEffect to reset when the open prop changes, covering parent-driven closes.

ui/client/src/components/agentic-ai/templates/CreateTemplateDialog.tsx
Lina-AbuYousef: Already fixed.

multi-agent/run/fixtures/seeder.py
Lina-AbuYousef: exactly, the user could have the yaml of the pre defined one as example and could use them with updated stuff. 
The UI checks admin status via the useAdminAccess() hook, which calls checkAdminAccess(userId) on the backend. The isAdmin flag is passed to the catalog and card components to conditionally show admin controls. On the server side, I've now also added @require_admin_access to both endpoints as defense-in-depth.

multi-agent/lib/mas/templates/repository/repository.py
Lina-AbuYousef: Correct -- there's currently no retention policy for soft-deleted templates and no template versioning beyond the metadata version string (which is display-only). Soft-deleted rows persist in MongoDB indefinitely. For now the volume is very low (only fixture templates + admin-created ones), so this isn't urgent, but we should track both as follow-up items:

multi-agent/run/fixtures/seeder.py
Lina-AbuYousef: fair point, fixed.

multi-agent/run/fixtures/seeder.py
```diff
@@ -0,0 +1,83 @@
...
+
+    yaml_files = sorted(FIXTURES_DIR.glob("*.yml")) + sorted(FIXTURES_DIR.glob("*.yaml"))
+    if not yaml_files:
+        logger.info("No YAML fixtures found in %s", FIXTURES_DIR)
+        return 0
+
+    inserted = 0
+    for path in yaml_files:
+        try:
+            with open(path, "r") as f:
```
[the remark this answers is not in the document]
Lina-AbuYousef: fair, fixed

Lina-AbuYousef:
@coderabbitai Please verify that the three fixture templates in `multi-agent/run/fixtures/templates/` match the following requirements from the Jira ticket:

**Template 1: "The L1/L2 SRE Auto-Medic"**
- Agents: OpenShift SRE, RHEL SRE, Jira, Reporter
- Scenario: Pod crash → inspect logs (OpenShift agent) → check host CPU/Memory (RHEL agent) → create Jira ticket with root-cause hypothesis → produce incident report
- All provider placeholders are configurable (no hard-coded URLs)

**Template 2: "The Autonomous Product Owner"**
- Agents: Confluence, Jira, Git
- Scenario: Read PRD from Confluence → inspect repo structure (Git agent) → create Jira Epics and Stories → produce sprint backlog report
- All provider placeholders are configurable

**Template 3: "The Continuous Compliance Auditor"**
- Agents: Git, Google Workspace, Reporter
- Scenario: Scan PRs in Git → cross-reference against security policies → generate compliance report
- All provider placeholders are configurable

For each template, check:
1. Does the blueprint define the correct agents with the right tools/providers?
2. Does the orchestrator's system prompt match the described scenario and execution flow?
3. Are all credentials and URLs exposed as placeholders (no hard-coded secrets or environment-specific URLs)?
4. Is the plan structure (steps, dependencies, branches) consistent with the orchestrator's described strategy?

multi-agent/run/fixtures/templates/blueprint_autonomous_product_owner.yml
```diff
@@ -0,0 +1,554 @@
...
+  # Providers
+  # ----------------------------
+  providers:
+    - rid: confluence_mcp_rid
+      name: Confluence MCP
+      type: mcp_server
+      config:
+        type: mcp_server
+        bearer_token: PLACEHOLDER_ATLASSIAN_PAT
+        mcp_url: https://mcp-atlassian-tag-ai--runtime-int.apps.stc-ai-e1-pp.imap.p1.openshiftapps.com/mcp
```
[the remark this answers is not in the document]
Lina-AbuYousef: this is intentional suggest for the users in templates to use this mcps, but they can change it later

multi-agent/run/fixtures/templates/blueprint_continuous_compliance_auditor.yml
```diff
@@ -0,0 +1,503 @@
...
+  # Providers
+  # ----------------------------
+  providers:
+    - rid: git_mcp_rid
+      name: Git MCP
+      type: mcp_server
+      config:
+        type: mcp_server
+        bearer_token: PLACEHOLDER_GIT_PAT
+        mcp_url: https://api.githubcopilot.com/mcp/
```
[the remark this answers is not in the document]
Lina-AbuYousef: sam comment

multi-agent/bootstrap/container.py
```diff
@@ -297,8 +297,19 @@ def __init__(self, cfg: AppConfig):
...
         )
 
+        self._seed_templates()
+
         self._initialized = True
 
+    def _seed_templates(self):
+        """Load YAML template fixtures into the database if not already present."""
+        from run.fixtures.seeder import seed_templates
+        logger = logging.getLogger(__name__)
```
Nirsisr: Lina's bring this import to top of the file

multi-agent/bootstrap/container.py
```diff
@@ -297,8 +297,19 @@ def __init__(self, cfg: AppConfig):
             cfg, self.session_repo, self.identity_provider
         )
 
+        self._seed_templates()
+
         self._initialized = True
 
+    def _seed_templates(self):
+        """Load YAML template fixtures into the database if not already present."""
+        from run.fixtures.seeder import seed_templates
```
Nirsisr: imports on top of the file, we try now to be aligned on that

multi-agent/lib/mas/templates/repository/repository.py
```diff
@@ -40,9 +40,11 @@ def update(self, template: Template) -> bool:
...
         
-        Returns True if a document was removed.
+        Returns True if a document was marked deleted.
+        The document stays in the database so the seeder
+        knows not to re-insert fixture templates.
```
Nirsisr, line 47: @Lina-AbuYousef Reminder to open a new ticket for retention in the DB for _deleted `True` elements, for clean the DB in a while.

Nirsisr: Ticket should cover not only templates area, but all the collections elements we save with _deleted (recently we have been added a lot like this)

multi-agent/run/fixtures/seeder.py
```diff
@@ -0,0 +1,83 @@
...
+    TemplateMetadata,
+)
+from mas.templates.repository.repository import TemplateRepository
+
+logger = logging.getLogger(__name__)
+
+FIXTURES_DIR = Path(__file__).parent / "templates"
+
+
+def seed_templates(repository: TemplateRepository) -> int:
```
Nirsisr, line 29: @Lina-AbuYousef How does it work exactly, admin can upload a new YML from the GUI which we want to insert to our DB, here under `templates` you have 3 pre-defined YML for dedicated templates, when user add from the GUI you add them to the DB automatically? 

If so, I recommend moving the 3 .yml templates from this folder to an external resources/repo like we are doing with SECRETS etc. Since, from one hand users can upload from the GUI directly to MONGO and from the other hand we have in the repo only 3 pre-defined. 

Like if user add from the GUI, it only reside on Mongo right, there is no automatic process to write the new .yml into our repo I guess

Nirsisr: Like the code you wrote here seems like a SCRIPT running one time to upload those PRE-DEFINED ymls to our DB represent TEMPLATES, that's why I prefer this code being called from a script file instead from here.

If we only need it once in the entire life-cycle of our code, why do we need this FIXTURE here?

Assuming we add a new YML someone I guess can run the script, not sure it should be part of our code. Since probably it's something that will take affect only once in a few months when user will add .yml to the TEMPLATES folder.

Nirsisr (changes requested):
Putting few notes, overall looking good @Lina-AbuYousef 10x. 

I will checkout to the branch to see how it looks in actions and will add comments if I have.

Nirsisr:
**Following our call here's what we want to adjust ATM:**
1) CSS adjustments with drak-mode.
2) YML validation check (used what we have already implemented (leverage it as part of the new `create teamplate` modal)
3) User responsibility to check the template he added with the JSON placeholders which might lead to issues.
4) Add ENV_VAR for admin_users @shanitzvii add to localhost.

multi-agent/run/fixtures/seeder.py
Lina-AbuYousef: The _seed_templates() call was removed from bootstrap/container.py. Seeding is now done via scripts/seed_templates.py — a one-time script you run manually whenever fixtures change

Lina-AbuYousef: The 3 .yml files under run/fixtures/templates/ are our source of truth for the pre-defined templates. They live in the repo for a few reasons:
Version control & review — any change to a pre-defined template goes through a PR, so the team can review it before it hits the DB. If they lived in an external resource, changes would bypass code review.
Reproducibility — anyone spinning up a fresh environment (new dev, new staging cluster, disaster recovery) can run the seed script and get a consistent baseline. Without the fixtures in the repo, you'd need to know where the external source is and how to fetch from it.

multi-agent/bootstrap/container.py
Lina-AbuYousef: done

multi-agent/bootstrap/container.py
Lina-AbuYousef: same

multi-agent/lib/mas/templates/repository/repository.py
Nirsisr: Resolving Linas. double-check we have this ticket in JIRA @Lina-AbuYousef

multi-agent/run/fixtures/seeder.py
Nirsisr: @Lina-AbuYousef Great that we have the script.

Regard the 3 YML explanation I'm not completely agree but it's ok. `Reproducibility` what about mongo persistence? Assuming different admins added templates so we have now 8 Templates, it means 5 are only on Mongo so if mongo is dropped we lost the entire 5 but only those 3 Fixtures will be always there since they are pre-defined. That's ok but I wonder why those 3 consider as special.

So if Nir want to add a template which he thinks it's GOLD, better for me add it to the REPO instead of using your `Add Template` feature since if i'm using the Add it might be gone .. coupled to the Mongo.
If that's the case why do we need the `Add Template` at all, that's where my point.

Anyway you can resolve, up to you i'm good with it both.

Lina-AbuYousef: yes, like the idea was that we have 3 pre defined ones, and of course adding new will only be in mongo
you can use the endpoint ,, but for thoso who feels more easy using the ui they can use the ui "like omri"

Nirsisr (changes requested):
@Lina-AbuYousef Closed second cycle, under `Create Template` page please disable the `Create Template` btn when the validation of the YML is not finished successfully. 

CSS is still not the same as we have in Inventory. it's Dark background with dark bg for the fields too on your feature, we have light bg for the fiels in Inventory elements, it's mot completely aligned but if you like this CSS I can accept it (since it's only for admins, in this case i'm not going to insist for consistency).

@odaiodeh If you can CR it too it's better, if not we will merge it.

Nirsisr (dismissed):

multi-agent/lib/mas/templates/repository/repository.py
sfiresht: no problem, resolving, please make sure this gets a followup item.

sfiresht (dismissed):
nice & straight forward, double like.

Nirsisr (dismissed):

sfiresht (dismissed):

Nirsisr (approved):

sfiresht (approved):

PR #135, opened 2026-03-29
UnifAI Workflow Execution Script

Nirsisr:
UnifAI Workflow Execution Script

================================

This script executes a workflow session on the UnifAI platform via API calls. It creates a session from a blueprint, submits a user prompt, polls for completion, and returns the final answer generated by the workflow.

```
(venv) bash-5.1$ python scripts/execution_workflow.py --help
usage: execution_workflow.py [-h] [-b BLUEPRINT_ID] [-u USER] [-q QUESTION] [-p POLLING_INTERVAL] [--base-url BASE_URL]

Execute a UnifAI workflow session and return the final answer.

options:
  -h, --help            show this help message and exit
  -b BLUEPRINT_ID, --blueprint-id BLUEPRINT_ID
                        Blueprint ID to execute (env: BLUEPRINT_ID)
  -u USER, --user USER  User shortcut, e.g., 'ocohen' (env: USER_SHORTCUT)
  -q QUESTION, --question QUESTION
                        Prompt/question for the workflow (env: USER_QUESTION)
  -p POLLING_INTERVAL, --polling-interval POLLING_INTERVAL
                        Status polling interval in seconds (5-60, default: 10)
  --base-url BASE_URL   UnifAI base URL (env: UNIFAI_BASE_URL)

Examples:
  python execution_workflow.py --blueprint-id abc-123 --user ocohen --question "Hello world"
  python execution_workflow.py -b abc-123 -u ocohen -q "Hello" --polling-interval 5
```


scripts/execution_workflow.py
```diff
@@ -0,0 +1,474 @@
...
+
+        is_active = status.get("is_active", False)
+        event_count = status.get("event_count", 0)
+        
+        if is_active:
+            seen_active = True
+        
+        print(f"      Poll #{poll_count}: events={event_count}, active={is_active}")
+        
+        if not is_active:
```
MayaCrmi, line 296: how does this work? how many times do we wait for is_active to make sure the script's not stuck?

scripts/execution_workflow.py
```diff
@@ -0,0 +1,474 @@
...
+
+# Suppress SSL warnings for internal routes
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+# =============================================================================
+# Configuration
+# =============================================================================
+
+DEFAULT_BASE_URL = "https://unifai-ui-tag-ai--pipeline.apps.stc-ai-e1-prod.rtc9.p1.openshiftapps.com"
```
MayaCrmi, line 70: will this be changable for example for us testing on PP?

scripts/execution_workflow.py
Nirsisr: We have a ENV argument called `polling_interval` controlled on the intervals we use to check is_active via API call, it's configurable by the user. Might be changed of course since the user who chose the script should know if it's workflows is long-running or short one. 

The script won't stuck, it's uses UnifAI APIs, we have protection layers in MAS if something stuck on RUNNING it will change status in the end. Than the script will consume it.

scripts/execution_workflow.py
Nirsisr: Fair question maya. it's configurable as well via the ENV param: UNIFAI_BASE_URL

scripts/execution_workflow.py
MayaCrmi: gotcha, nice!

scripts/execution_workflow.py
MayaCrmi: alrighty :)

sfiresht:
@Nirsisr how would you run this script once the multiagent is refactored to get cookies from the UI? in that case you'll have to find a way to use authentication somehow of get a token or somthing...

scripts/execution_workflow.py
```diff
@@ -0,0 +1,474 @@
+#!/usr/bin/env python3
```
sfiresht, line 1: @Nirsisr how would you run this script once the multiagent is refactored to get cookies from the UI? in that case you'll have to find a way to use authentication somehow of get a token or somthing..

sfiresht (commented):
d

sfiresht (approved):

Nirsisr:
> @Nirsisr how would you run this script once the multiagent is refactored to get cookies from the UI? in that case you'll have to find a way to use authentication somehow of get a token or somthing...

As we start the discussion privately, that's a great question.
Here is my vision: each script/CLI integration we currently have will start his flow, calling another script which validate the user first, than we hold the cookie/session_id in HAND for the entire lifecycle of the SCRIPT.

How to maintain it when session is terminated? We need to think of it.
Regarding cleaning SCRIPTS sessions that where created, we need to come up with an approach for those cases. Maybe some meta_data VAR called SCRIPTS that we will manage different strategy for TTL cleanup on those.

scripts/execution_workflow.py
Nirsisr: Yep saariko, agree, see my answer below. Each SCRIPT rely on USER identity will need to be REFACTORED.

yhabushi79 (approved):

odaiodeh (approved):

PR #143, opened 2026-04-12
GENIE-1504: Fix RAG Document Pipeline Transport Error

yhabushi79:
## Problem

`DoclingClient.post_file()` held a single HTTP connection open for the entire document conversion.
Long conversions (minutes) caused intermediate proxies / load balancers (OpenShift router, HAProxy)
to kill the idle connection, resulting in a `TransportError`.

## Fix — Async Submit + Poll + Fetch

Replaced the single long-lived call with three short-lived HTTP requests:

1. **Submit** (`POST /v1/convert/file/async`) — enqueues the file, returns a `task_id` immediately.
2. **Poll** (`GET /v1/status/poll/{task_id}`) — checks status every 10 s with fresh, short-lived requests. Tolerates up to 3 consecutive transient network failures.
3. **Fetch** (`GET /v1/result/{task_id}`) — retrieves the result once conversion succeeds.

## Flow

```
Before (broken):

  post_file() ──[ single long HTTP POST ]──▶ /v1/convert/file
                  ▲ connection idle for minutes
                  ▲ proxy kills it → TransportError


After (fixed):

  post_file()
    │
    ├─ _submit_file()
    │     POST /v1/convert/file/async  →  returns task_id  (fast)
    │
    ├─ _poll_until_done()
    │     loop every 10 s:
    │       GET /v1/status/poll/{task_id}
    │       ├─ status == "success"  →  break
    │       ├─ status == "failure"  →  raise error
    │       └─ transient error      →  retry (up to 3×)
    │
    └─ _fetch_result()
          GET /v1/result/{task_id}  →  returns conversion output  
```


MayaCrmi (dismissed):

global_utils/src/global_utils/docling/client.py
```diff
@@ -18,6 +23,9 @@
 
 logger = logging.getLogger(__name__)
 
+_POLL_INTERVAL_SECONDS = 10
```
sfiresht: don't we want this as params passed from helm values? or a config in the config file?

global_utils/src/global_utils/docling/client.py
```diff
@@ -46,16 +58,41 @@ def __init__(
...
-            timeout: Request timeout in seconds
+            timeout: Total wall-clock timeout in seconds for a full
+                     submit + poll + fetch cycle
         """
         self.base_url = base_url.rstrip('/')
         self.timeout = timeout
         self._client = httpx.Client(
             base_url=self.base_url,
-            timeout=httpx.Timeout(timeout),
+            timeout=httpx.Timeout(60),
```
sfiresht: don't we want this as params passed from helm values? or a config in the config file?

global_utils/src/global_utils/docling/client.py
sfiresht: resolving after checking with Yosi, having this in the config file is enough as no one really plays with it all the time.

global_utils/src/global_utils/docling/client.py
sfiresht: resolving after checking with Yosi, having this in the config file is enough as no one really plays with it all the time.

sfiresht (approved):

Nirsisr (approved):
Approved, confirmed with Yoss feature where tested (decided with Omri to not block the PR, 1 approval is accepted)

Nirsisr (commented):
@yhabushi79 Yoss do we cover the case if the pulling never end? Assuming something got stuck for some reason in docling and we never update the STATUS? Or we are making sure always after X min to change the status in CATCH mechanism under the docling implementation? 

Not sure what `time.monotonic()` stands for, i'm just worried we will have infinite loops over here if something got stuck on the docling side.

PR #145, opened 2026-04-14
GENIE-1151/story/name-duplication-logic

MayaCrmi:
currently, when a user creates or edits an inventory object and enters a name that already exists, he gets the feedback in an error toast after clicking save. by then, the form closes and all configuration he made is lost. this can be really frustrating for bigger objects where the user has spent time filling in urls, api keys, system messages, etc.

for this, i added real-time inline validation on the name field that checks against existing element names (case-insensitive) and displays an error message directly below the field. the save button is disabled while a duplicate name is detected. also, when editing an existing element, its own current name is excluded from the duplicate check.

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent duplicate element names in forms; error messages now display when a name already exists.

### DIFF
--- a/ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
+++ b/ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from "react";
+import React, { useState, useEffect, useCallback, useMemo } from "react";
 import {
   Dialog,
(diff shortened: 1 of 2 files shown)


sfiresht (approved):

Lina-AbuYousef (approved):

MayaCrmi:
pulling from main created some problems here so closing to reopen it

PR #37, opened 2025-12-22
GENIE-1067/STORY/create infra validation layer

odaiodeh:
### Summary

Added a complete validation system for validating element configurations, saved resources, and blueprints with network connectivity checks and dependency validation.

---

#### New API Endpoints

| Endpoint | Purpose |
|----------|---------|
| `POST /api/resources/resource.validate` | Validate saved resource + dependencies |
| `POST /api/resources/config.validate` | Validate config before saving |
| `POST /api/blueprints/blueprint.validate` | Validate all elements in saved blueprint |
| `POST /api/blueprints/draft.validate` | Validate blueprint draft before saving |

#### New Element Validators

| Element | Validator | Checks |
|---------|-----------|--------|
| `mcp_server` | `McpProviderValidator` | Connects to MCP endpoint, lists tools |
| `a2a_agent_node` | `A2AAgentNodeValidator` | Fetches agent card from A2A endpoint |
| `custom_agent_node` | `CustomAgentNodeValidator` | Verifies all dependencies are valid |

---

### Architecture

```
┌─────────────────────────────────────────────────────────────────────────┐
│                            API LAYER                                    │
│   /resources/resource.validate    /blueprints/blueprint.validate        │
│   /resources/config.validate      /blueprints/draft.validate            │
└───────────────────────────────────┬─────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────────┐
│                          SERVICE LAYER                                  │
│  ┌──────────────────────┐              ┌──────────────────────┐         │
│  │   ResourcesService   │              │   BlueprintService   │         │
│  │ • validate_resource  │              │ • validate_blueprint │         │
│  │ • validate_config    │              │ • validate_draft     │         │
│  └──────────┬───────────┘              └──────────┬───────────┘         │
│             │                                     │                     │
│             └──────────►┌────────────────────────┐◄──────────┘          │
│                         │ ElementValidationService│                     │
│                         │ • validate()            │                     │
│                         │ • validate_ordered()    │                     │
│                         └────────────┬────────────┘                     │
└──────────────────────────────────────┼──────────────────────────────────┘
                                       │
                                       ▼
┌─────────────────────────────────────────────────────────────────────────┐
│                        VALIDATOR LAYER                                  │
│   ┌─────────────────┐ ┌─────────────────┐ ┌─────────────────┐           │
│   │ McpProvider     │ │ A2AAgentNode    │ │ CustomAgent     │           │
[...]

PR #35, opened 2025-12-21
hexagonal Architecture Migration

Lina-AbuYousef:
# RAG Module: Hexagonal Architecture Migration

## 🎯 Summary

Migrated the backend data pipeline (`/backend`) to a new `/rag` module using **Hexagonal Architecture** (Ports & Adapters). This is a **design refactoring** — the business logic remains identical, but the code structure now follows clean architecture principles for better testability and maintainability.

## ⚠️ Important: Design Changes, Not Logic Changes

This PR does **not** change any business logic. All pipeline execution, validation, and storage operations work exactly as before. The changes are purely structural:

| Aspect | Before (`/backend`) | After (`/rag`) |
|--------|---------------------|----------------|
| Architecture | Mixed concerns | Hexagonal (Ports & Adapters) |
| Dependencies | Scattered imports | Centralized DI container |
| Testing | Mock internals | Inject mock ports |
| Coupling | Tight (DB in handlers) | Loose (via interfaces) |

---

### `domain/` — The Core (Pure Business Logic)

**What it contains:**
- **Models (Entities)**: `PipelineRecord`, `DataSource`, `VectorChunk`, `SearchResult`
- **Ports (Interfaces)**: Abstract classes defining contracts (`VectorRepository`, `PipelineRepository`, `SourcePipelinePort`)
- **Domain Processors**: Stateless transformers (`SlackProcessor`, `DocumentProcessor`)

**Principles:**
- ✅ Zero external dependencies (no Flask, no MongoDB, no Qdrant imports)
- ✅ Pure Python — can be tested without any infrastructure
- ✅ Defines "what" not "how" (interfaces, not implementations)

### `application/` — Use Cases (Orchestration)

**What it contains:**
- **Services**: `PipelineService`, `DataSourceService`, `RetrievalService`, `MonitoringService`
- **Pipeline Handlers**: `DocumentPipelineHandler`, `SlackPipelineHandler` (implement `SourcePipelinePort`)
- **Validators**: Validation pipeline for documents and Slack channels
- **Registration**: Source registration workflow

**Principles:**
- ✅ Depends only on `domain/` (never on `infrastructure/`)
- ✅ Orchestrates domain objects to fulfill use cases
- ✅ Receives dependencies via constructor injection (ports, not implementations)

### `infrastructure/` — Adapters (External Integrations)

**What it contains:**
- **Driven Adapters** (output): `MongoPipelineRepository`, `QdrantVectorRepository`, `SentenceTransformerEmbedder`
[...]

PR #34, opened 2025-12-18
Genie 1032/story/add overall statistics dashbaord

yhabushi79:
This branch adds a comprehensive Workflow Analytics Dashboard that provides system-wide statistics and insights. The feature includes both backend services and a React frontend dashboard.

Screenshot 2025-12-18 at 17 04 53

# Analytics Module

## Overview

The Analytics module provides **system-wide workflow statistics** for the admin dashboard. It is a new feature added to UnifAI that enables administrators to monitor workflow execution patterns, user activity, and blueprint usage across the entire platform.

This module follows the established architectural patterns in the `multi-agent` codebase:
- **Service-Repository Pattern** (like `ShareService` / `MongoShareRepository`)
- **Dependency Injection** via `AppContainer`
- **Pydantic DTOs** for type-safe data transfer
- **Abstract Repository** for database abstraction

### ASCII Architecture Diagram

```
┌─────────────────────────────────────────────────────────────────────────────────────┐
│                                    FRONTEND                                          │
│                                                                                      │
│  ┌──────────────────┐    ┌────────────────────┐    ┌─────────────────────────────┐  │
│  │  Analytics.tsx   │───►│  systemStats.ts    │───►│  GET /statistics/           │  │
│  │  (React Page)    │    │  (API Client)      │    │  stats.system.get           │  │
│  └──────────────────┘    └────────────────────┘    └─────────────────────────────┘  │
│         │                                                       │                    │
│         ▼                                                       │                    │
│  ┌──────────────────┐    ┌────────────────────┐                 │                    │
│  │  systemStats.ts  │    │  Components:       │                 │                    │
│  │  (TypeScript     │    │  - StatusBreakdown │                 │                    │
│  │   Interfaces)    │    │  - TopUsersChart   │                 │                    │
│  └──────────────────┘    │  - BlueprintsTable │                 │                    │
│                          └────────────────────┘                 │                    │
└─────────────────────────────────────────────────────────────────┼────────────────────┘
                                                                  │
                                                                  ▼
┌─────────────────────────────────────────────────────────────────────────────────────┐
│                                    BACKEND                                           │
│                                                                                      │
│  ┌─────────────────────────────────────────────────────────────────────────────┐    │
│  │                           API LAYER                                          │    │
│  │  ┌─────────────────────┐    ┌─────────────────────────────────────────────┐ │    │
│  │  │  statistics.py      │    │  decorators.py                              │ │    │
│  │  │  /stats.system.get  │◄───│  @require_admin_access                      │ │    │
[...]

PR #147, opened 2026-04-15
Genie 1501/Add local development guide

shanitzvii:
Added a comprehensive local development guide and two helper scripts to set up and run the entire UnifAI stack on a developer's machine - either the full stack or a single service.

Changes Summary:
1) local-development/LOCAL_DEV_GUIDE.md - End-to-end guide covering both development modes: prerequisites, venv setup for each service, helper script usage, typical per-service workflows, and troubleshooting known issues.

2) local-development/apply-local-dev-changes.py - Patches config files so all services bind and discover each other on localhost. Supports --setup-venv to auto-create virtual environments

3) local-development/start-unifai-dev.sh - Single entry point for two modes:
- Full-stack (default): applies config patches, starts all infrastructure containers (MongoDB, RabbitMQ, Qdrant, Redis, Temporal), and opens a tmux session with all 5 services + 2 workers.
- Single-service (--service <name>): starts only the containers the service needs and runs it in the foreground with live auto-reload.

4) local-development/start-infra.sh — Standalone utility to manage infrastructure containers: start per-service dependencies (--for <name>), cherry-pick specific containers, check status, or stop all.

5) README.md - Replaced inline Quick Start with a link to the new guide.

6) .github/CODEOWNERS - Added the new library loval-development and set code owners

## Summary by CodeRabbit

* **Chores**
  * Updated repository code ownership to add an additional owner across relevant areas.
  * Made the local development server host and port configurable via environment variables to simplify local setup and environment parity.
  * Added a new local proxy route to forward development API requests to the backend, easing integration and testing during development.


local-development/apply-local-dev-changes.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/env python3
```
sfiresht: as discussed, this script might cause a problem where users update the configuration and push the "updated" on to the repo, > causing deployment to fail due to localized config.

I suggest to change this script so all it does is against env var files  (or some equivalenet) that way the users will only update a file that is used locally >  no danger to deployment.

in addition if this script remains as is I'd like to add a very clear notification in the guis that all configuration changed MUST be reverted prior to pushing in order to prevent real deployment.

local-development/start-unifai-dev.sh
```diff
@@ -0,0 +1,246 @@
...
+# Python interpreter — override with: export UNIFAI_PYTHON=python3.12
+if [ -n "$UNIFAI_PYTHON" ]; then
+    if ! command -v "$UNIFAI_PYTHON" &> /dev/null; then
+        echo "❌ UNIFAI_PYTHON is set to '$UNIFAI_PYTHON' but that command was not found."
+        exit 1
+    fi
+    PYTHON="$UNIFAI_PYTHON"
+else
+    PYTHON=""
+    for candidate in python3.11 python3.12 python3; do
```
sfiresht: will it work for python 3.13?

local-development/start-unifai-dev.sh
```diff
@@ -0,0 +1,246 @@
+#!/bin/bash
```
sfiresht: as you've already found version 3.14 isn't supported due to pydantic use of PyO3.

maybe a notification about versions support should be added .

local-development/apply-local-dev-changes.py
shanitzvii: I've added env files generation where I could - 
  - rag/.env
  - ui/.env.local
  - shared-resources/sso-backend/.env

I kept the minimal patches within the script work:
Bind host 0.0.0.0 in 3 entrypoints
/api3 proxy route injection into vite.config.ts - as @Nirsisr told me to keep it for now as a local patch.

Added warnings about patches in the guide (right under section 4. Running the Development Environment)

local-development/start-unifai-dev.sh
shanitzvii: canged validation to accept  python 3.11-3.13 versions only

MayaCrmi:
it looks good, really thorough! assuming saar and you are going to take a look into the more sensitive parts it seems promising. the best way to make sure it works is making someone new use it. i know Odai sent it out this past weekend to someone so maybe we can hear his thoughts? @odaiodeh

sfiresht (dismissed):
looks cool!

local-development/apply-local-dev-changes.py
sfiresht: fixed, resolving

sfiresht (dismissed):

odaiodeh (approved):

sfiresht (dismissed):

MayaCrmi (dismissed):
Glad to hear it worked for new timers!!

yhabushi79 (dismissed):

odaiodeh (changes requested):
We have secrets keys of sso embedded int he code, and now its in this branch in public repo, this exposing our key to the public

local-development/apply-local-dev-changes.py
```diff
@@ -0,0 +1,457 @@
...
+"""
+Script to apply local development configuration changes.
+Modifies configuration files so all services can run side-by-side on localhost.
+
+Port layout after patching:
+  RAG           → 13457  (changed from 13456 to avoid SSO collision)
+  SSO Backend   → 13456
+  Multi-Agent   → 8002   (unchanged)
+  Backend       → 8005   (unchanged)
+  UI (Vite)     → 5000   (changed from 5173)
```
Nirsisr: Please remove from here the (unchanged command) those are for us the developers no need to have it in the repo (no need for backward compatibility)

shanitzvii:
both comments were handled @odaiodeh @Nirsisr

local-development/start-unifai-dev.sh
```diff
@@ -0,0 +1,526 @@
...
+
+SERVICE=""
+SETUP_VENV=false
+SKIP_PATCH=false
+FORCE_ENV=false
+DESTROY=false
+
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --destroy|--destory)
```
yhabushi79: typo - its my bad. since it going to other teams we should be correct.

local-development/apply-local-dev-changes.py
```diff
@@ -0,0 +1,454 @@
...
+def _is_unifai_root(path: Path) -> bool:
+    path = path.resolve()
+    return (
+        (path / "rag" / "config" / "app_config.py").is_file()
+        and (path / "ui" / "vite.config.ts").is_file()
+        and (path / "global_utils" / "src" / "global_utils" / "config" / "config.py").is_file()
+    )
+
+
+def resolve_unifai_root() -> Path:
```
yhabushi79: when you think about it. you know already the project structure.
we can always say to user that they should run the script from a specific location.

local-development/start-unifai-dev.sh
```diff
@@ -0,0 +1,526 @@
...
+    echo ""
+
+    case "$SERVICE" in
+        backend)
+            kill_port 8005
+            check_venv "$ROOT_DIR/backend/venv"
+            check_venv_python "$ROOT_DIR/backend/venv"
+            cd "$ROOT_DIR/backend"
+            source venv/bin/activate
+            set -a && source .env 2>/dev/null; set +a
```
yhabushi79: when i execute the script i noticed that only rag has .env file created.

ui/vite.config.ts
```diff
@@ -43,6 +46,12 @@ export default defineConfig({
         rewrite: (path) => path.replace(/^\/api2/, '/api'), // This rewrites /api2 to nothing
         // secure: false, // Only needed if this target is HTTPS and you have SSL issues
```
sfiresht, line 47: any reason not to include api3 as well? this one is for the sso module.

local-development/start-unifai-dev.sh
```diff
@@ -0,0 +1,526 @@
...
+            kill_port 8002
+            check_venv "$ROOT_DIR/multi-agent/venv"
+            check_venv_python "$ROOT_DIR/multi-agent/venv"
+            cd "$ROOT_DIR/multi-agent"
+            source venv/bin/activate
+            export HOSTNAME=0.0.0.0
+            exec mas api dev
+            ;;
+
+        sso)
```
sfiresht: depending on whether my branch will be pushed prior to yours this might be changed to fit the new identity service.

local-development/apply-local-dev-changes.py
```diff
@@ -0,0 +1,454 @@
...
+
+    # ── 6. shared-resources/sso-backend/app.py ────────────────────────────
+    #    Bind to 0.0.0.0 explicitly
+    print("  📝 Updating shared-resources/sso-backend/app.py...")
+    file_path = p("shared-resources", "sso-backend", "app.py")
+    content = file_path.read_text()
+    content = content.replace(
+        "app.run(host=config.hostname_local, port=config.port, debug=True)",
+        'app.run("0.0.0.0", port=config.port, debug=True)',
+    )
```
sfiresht: depending on whether my branch will be pushed prior to yours this might be changed to fit the new identity service.

Nirsisr:
> both comments were handled @odaiodeh @Nirsisr

Thanks shani, as discussed F2F suggested few improvements for the new local-dev folder (you can take some inspiration from hexagonal structure we are maintaining in other services, global_utils/config abstraction we are leveraging to define config files across the repo, same concept need to be done with local-dev folder putting there all the general configurations).

Another concept is abstraction e.g. for VENV_Definition that each service (rag, mas, ui) will need to implement in his own class the `venv_build` function for example. The .py file will consume from all the services the build method along other methods that will be implemented per service which will clarify how to run the specific tmux session per the relevant service.

local-development/start-infra.sh
```diff
@@ -0,0 +1,238 @@
...
+    [sso]=""
+    [ui]=""
+    [celery-worker]="mongo rabbitmq qdrant"
+    [temporal-worker]="mongo redis temporal"
+)
+
+# ── Detect container runtime ─────────────────────────────────────────────────
+
+detect_runtime() {
+    if command -v podman &> /dev/null && podman info &> /dev/null 2>&1; then
```
yhabushi79: i think there might be an issue when you have an existing container on the vm/localdev
i encountered an issue several time when running locally

`yhabushi@yhabushi-mac local-development % zsh start-infra.sh
Using container runtime: podman

Starting infrastructure: mongo qdrant rabbitmq redis temporal

  ⊕ Creating MongoDB container…
  ⚠ Failed to create MongoDB.
yhabushi@yhabushi-mac local-development % `

the error is streightforward but the user will not have the abiility to understand why it fails becuase we pass the std/err to /dev/null

odaiodeh:
Hi @shanitzvii here is a review of this PR, it first shows the violations and issues in the code, then it shows a suggestion of proposed architecture.
# PR Review: `GENIE-1501_local_dev_guide`

## Overview

This PR introduces a local development toolchain for UnifAI: a comprehensive guide (`LOCAL_DEV_GUIDE.md`), a configuration patcher (`apply-local-dev-changes.py`), an infrastructure manager (`start-infra.sh`), a dev orchestrator (`start-unifai-dev.sh`), Vite config changes, and supporting updates to `CODEOWNERS` and `README.md`.

The effort here is substantial and the documentation is excellent. The tooling works and solves a real pain point. This review focuses on architectural concerns that will become costly if not addressed now, while the codebase is still young.

---

## Table of Contents

- [1. Code Smells](#1-code-smells)
  - [1.1 Duplicated Logic Across Files](#11-duplicated-logic-across-files)
  - [1.2 God Script](#12-god-script-start-unifai-devsh)
  - [1.3 Fragile String-Replacement Patching](#13-fragile-string-replacement-patching)
  - [1.4 Source Patching That Must Be Manually Reverted](#14-source-patching-that-must-be-manually-reverted)
  - [1.5 Committed Code Broken Without Tooling](#15-committed-code-broken-without-tooling)
  - [1.6 Script Fighting Its Own Config](#16-script-fighting-its-own-config)
  - [1.7 Scattered Service Knowledge](#17-scattered-service-knowledge-shotgun-surgery)
  - [1.8 Inconsistent Proxy Route Patterns](#18-inconsistent-proxy-route-patterns)
  - [1.9 Unused Module-Level Variable](#19-unused-module-level-variable)
- [2. SOLID Violations](#2-solid-violations)
  - [2.1 Single Responsibility Principle](#21-single-responsibility-principle)
  - [2.2 Open/Closed Principle](#22-openclosed-principle)
  - [2.3 Interface Segregation Principle](#23-interface-segregation-principle)
  - [2.4 Dependency Inversion Principle](#24-dependency-inversion-principle)
- [3. Proposed Architecture](#3-proposed-architecture)
  - [3.1 Eliminate Source Patching Entirely](#31-eliminate-source-patching-entirely)
  - [3.2 Service Registry (Single Source of Truth)](#32-service-registry-single-source-of-truth)
  - [3.3 Hexagonal Tooling Architecture](#33-hexagonal-tooling-architecture)
  - [3.4 Composable tmux Interface](#34-composable-tmux-interface)
  - [3.5 Recovery and Observability](#35-recovery-and-observability)
  - [3.6 Proposed CLI](#36-proposed-cli)
  - [3.7 What "Adding a New Service" Looks Like](#37-what-adding-a-new-service-looks-like)
- [4. Priority Matrix](#4-priority-matrix)

---

## 1. Code Smells

### 1.1 Duplicated Logic Across Files

The `ensure_container()` function is implemented **twice** with different signatures:

**In `start-infra.sh` (lines 103-130)** — clean, declarative:
```bash
ensure_container() {
    local name="$1"
    local label="${CONTAINER_LABELS[$name]}"
    local image="${CONTAINER_IMAGES[$name]}"
    # reads from declarative associative arrays
}
```

**In `start-unifai-dev.sh` (lines 362-381)** — inline, different interface:
```bash
ensure_container() {
    local name="$1"
    local label="$2"
    shift 2
    local run_args=("$@")
    # takes all args inline, no shared data source
}
```

Full-stack mode in `start-unifai-dev.sh` **does not call `start-infra.sh`** at all — it reimplements container management from scratch. Single-service mode correctly delegates to `start-infra.sh` (line 186), but full-stack mode ignores it.

The same service knowledge (ports, directories, commands, infra dependencies) is duplicated across all three files:

```
                    start-unifai-dev.sh     start-infra.sh     apply-local-dev-changes.py
                    ───────────────────     ──────────────     ──────────────────────────
Port numbers         lines 424, 244-289     lines 37-43        lines 291-342
Directories          lines 247-308          (none)             lines 165-218
Infra dependencies   lines 411-416          lines 68-76        (none)
Venv setup           (delegates)            (none)             lines 163-218
Container mgmt       lines 362-381          lines 103-130      (none)
```

**Risk**: These will drift. When someone changes a port or adds a container, they'll update one file and forget the others.

---

### 1.2 God Script: `start-unifai-dev.sh`

At 526 lines, this script handles **7 distinct responsibilities**:

1. CLI argument parsing (lines 30-89)
2. Python version detection and validation (lines 91-129)
3. Virtual environment setup delegation (lines 144-158)
4. Source-file and `.env` patching delegation (lines 160-172)
5. Infrastructure container management (lines 326-419) — duplicated from `start-infra.sh`
6. Port cleanup / process killing (lines 191-199, 423-431)
7. tmux session orchestration with hardcoded layout (lines 433-524)

Each of these is a separate concern that changes for different reasons. A port change touches concern 6. A new service touches concerns 1, 5, 6, and 7. A tmux layout change touches only concern 7. They should not be in the same file.

---

### 1.3 Fragile String-Replacement Patching

`apply-local-dev-changes.py` uses raw `str.replace()` with no validation:

```python
# line 281-284
content = content.replace(
    "host=config.hostname_local,",
    'host="0.0.0.0",',
)
```

**Problems**:
- If someone reformats the source to `host = config.hostname_local,` (with spaces), the replace silently does nothing — no warning, no error.
- If the target string has already been replaced (idempotency), the script still rewrites the file.
- There is no check that the replacement actually happened. The script will always report `modified.append("rag/bootstrap/flask_app.py")` whether or not the string was found.

Similarly, the regex injection for the `/api3` proxy route (lines 320-328):

```python
api2_pattern = r"(      '/api2': \{.*?\n      \},)"
```

This regex depends on exact indentation (6 spaces). Any formatting change to `vite.config.ts` (e.g., from Prettier, or a tab-vs-space change) will silently break the injection.

---

### 1.4 Source Patching That Must Be Manually Reverted

The setup script patches 4 source files that are **tracked by git**:

- `rag/bootstrap/flask_app.py`
- `backend/run/dev.py`
- `shared-resources/sso-backend/app.py`
- `ui/vite.config.ts`

The guide warns developers to manually run `git checkout` on these files before pushing. There is no automated safety net (no pre-commit hook, no pre-push hook, no CI check). If a developer forgets, the patched code (hardcoded `host="0.0.0.0"`, injected proxy routes) goes to production.

This is the highest-risk smell in the PR.

---

### 1.5 Committed Code Broken Without Tooling

The committed `vite.config.ts` now uses environment variables for proxy targets with **no fallbacks**:

```typescript
// line 36 — committed code
target: process.env.RAG_HOST,        // → undefined without .env.local
// line 44
target: process.env.MULTIAGENT_HOST, // → undefined without .env.local
// line 50
target: process.env.BACKEND_HOST,    // → undefined without .env.local
```

If a developer runs `pnpm dev` without first running `apply-local-dev-changes.py`, all three proxy targets resolve to `undefined` and every API call fails silently. The **original code** had hardcoded URLs (`'http://127.0.0.1:13457'`) that worked standalone — those were removed.

---

### 1.6 Script Fighting Its Own Config

Looking at the actual service configs, all three Python services **already default** `hostname_local` to `"0.0.0.0"`:

```python
# backend/config/app_config.py
hostname_local: str = "0.0.0.0"    # already 0.0.0.0

# rag/config/app_config.py
hostname_local: str = "0.0.0.0"    # already 0.0.0.0

# shared-resources/sso-backend/config/app_config.py
hostname_local: str = "0.0.0.0"    # already 0.0.0.0
```

The patch script does this:
1. **Generates `.env`** with `hostname_local=127.0.0.1` → overrides the default `0.0.0.0` to `127.0.0.1`
2. **Patches the source** to hardcode `host="0.0.0.0"` → bypasses the `.env` it just wrote

The script is fighting itself: step 1 sets `hostname_local=127.0.0.1`, step 2 ignores it. The root cause is that `hostname_local` is used for two separate concepts:
- **Bind address**: what the server listens on (needs `0.0.0.0`)
- **Connect address**: what other services use to reach this service (needs `127.0.0.1`)

---

### 1.7 Scattered Service Knowledge (Shotgun Surgery)

Adding a new service (e.g., `analytics`) requires editing **7+ files across 15+ locations**:

| # | File | What to change |
|---|------|----------------|
| 1 | `start-unifai-dev.sh` | Add `case` block in single-service mode (~line 242) |
| 2 | `start-unifai-dev.sh` | Add port to `PORTS` string (~line 424) |
| 3 | `start-unifai-dev.sh` | Add `verify_venv` call (~line 406) |
| 4 | `start-unifai-dev.sh` | Add tmux pane creation (~line 446+) |
| 5 | `start-unifai-dev.sh` | Add `ensure_container` calls (~line 411+) |
| 6 | `start-infra.sh` | Add to `SERVICE_INFRA` mapping (~line 68) |
| 7 | `apply-local-dev-changes.py` | Add venv table entry (~line 163) |
| 8 | `apply-local-dev-changes.py` | Add `.env` generation (~line 265+) |
| 9 | `apply-local-dev-changes.py` | Add source patch if needed |
| 10 | `LOCAL_DEV_GUIDE.md` | Update 5+ tables |
| 11 | `vite.config.ts` | Add proxy route if it has an API |

This is the [Shotgun Surgery](https://refactoring.guru/smells/shotgun-surgery) smell — one logical change requires many scattered edits.

---

### 1.8 Inconsistent Proxy Route Patterns

```
PROXY ROUTE        IN COMMITTED CODE?    TARGET SOURCE
──────────────     ──────────────────    ─────────────────────
/api1              yes                   process.env.RAG_HOST (no fallback)
/api2              yes                   process.env.MULTIAGENT_HOST (no fallback)
/api3              NO — injected at      process.env.SSO_HOST
                   runtime by regex
/api4              yes                   process.env.BACKEND_HOST (no fallback)
```

Three different patterns for the same concept: `/api1` and `/api2` were pre-existing (but had fallback URLs removed), `/api3` is runtime-injected, `/api4` is newly committed. A reader of `vite.config.ts` sees `/api1`, `/api2`, `/api4` but not `/api3` — there's no indication that another route gets injected.

---

### 1.9 Unused Module-Level Variable

```python
# apply-local-dev-changes.py line 160
ALL_VENV_SERVICES: dict[str, tuple[str, list[list[str]]]] = {}
```

This is declared at module level but never used anywhere. `_build_venv_table()` returns a new dict each time without referencing `ALL_VENV_SERVICES`.

---

## 2. SOLID Violations

### 2.1 Single Responsibility Principle

**`start-unifai-dev.sh`** has 7 responsibilities (see [1.2](#12-god-script-start-unifai-devsh)). Each changes for independent reasons.

**`apply-local-dev-changes.py`** has 3 responsibilities:
1. Venv creation (`setup_venvs`) — changes when a service's dependencies change
2. `.env` generation — changes when config values change
3. Source-file patching — changes when upstream source code changes

These should be separate modules/scripts.

---

### 2.2 Open/Closed Principle

The system is **closed for extension**. There is no way to add a service without modifying internal code.

Service definitions are embedded in `case` statements, inline function calls, and associative arrays across multiple files. There is no external configuration — the code **is** the configuration.

A compliant design would use a declarative service registry that the scripts read from, so adding a service means adding data, not changing code.

---

### 2.3 Interface Segregation Principle

The developer interface is a fat binary choice:

```
OPTION A:  --service backend     →  ONE service, foreground, no tmux
OPTION B:  (no flags)            →  ALL 7 services, fixed tmux layout
```

There is no way to:
- Start 2-3 services together in tmux
- Customize which services go in which tmux window
- Start just the workers
- Start just the API services

The interface forces the "all" contract on developers who only need a subset.

---

### 2.4 Dependency Inversion Principle

The high-level orchestrator (`start-unifai-dev.sh`) depends directly on low-level details:

```
HARDCODED IN THE ORCHESTRATOR:
──────────────────────────────
  Port numbers:    8005, 13457, 8002, 13456, 5000
  Directories:     "backend", "rag", "multi-agent", ...
  Venv paths:      "$ROOT_DIR/backend/venv", ...
  Launch commands: "$PYTHON -m run.dev", "mas api dev", ...
  Env files:       ".env", ".env.local"
  Tmux geometry:   split-window -h, split-window -v, ...
  Container args:  -p 27017:27017, mongo:latest, ...
```

The orchestrator should depend on abstractions (a registry, a container runtime interface), not on concrete ports, paths, and commands.

---

## 3. Proposed Architecture

### 3.1 Eliminate Source Patching Entirely

This is the single highest-impact change. The Pydantic `SharedConfig` system already supports `.env` files, environment variables, YAML, and JSON. The services already default `hostname_local` to `"0.0.0.0"`. The source patching exists because one field (`hostname_local`) is used for two different concepts.

**Fix: Add `bind_host` to `SharedConfig`**

```python
# global_utils/src/global_utils/config/config.py
class SharedConfig(BaseSettings):
    bind_host: str = "0.0.0.0"     # ← NEW: what servers listen on
    # ... existing fields ...
```

**Update each service entry point (one line each):**

```python
# rag/bootstrap/flask_app.py — BEFORE:
application.run(host=config.hostname_local, port=config.port, debug=True)
# rag/bootstrap/flask_app.py — AFTER:
application.run(host=config.bind_host, port=config.port, debug=True)

# backend/run/dev.py — same change
# shared-resources/sso-backend/app.py — same change
```

**The `.env` files (already gitignored) handle the rest:**

```env
# rag/.env
hostname_local=127.0.0.1   # what clients use to connect to this service
port=13457                  # avoid SSO collision
# bind_host=0.0.0.0        # already the default — not needed
```

**Result:**
- Zero source-file patching
- Zero risk of forgetting to revert before push
- No need for pre-push hooks
- The config system handles everything natively
- `apply-local-dev-changes.py` shrinks dramatically — it only generates `.env` files

**Fix `vite.config.ts` the same way** — commit all 4 proxy routes with env-var targets AND safe fallbacks:

```typescript
'/api1': {
  target: process.env.RAG_HOST || 'http://127.0.0.1:13457',
  // ...
},
'/api3': {  // committed, not injected at runtime
  target: process.env.SSO_HOST || 'http://127.0.0.1:13456',
  // ...
},
```

Now `pnpm dev` works standalone. The `.env.local` can override if needed, but the defaults are correct.

---

### 3.2 Service Registry (Single Source of Truth)

Replace scattered service knowledge with a single declarative file:

```yaml
# local-development/services.yaml

infrastructure:
  mongo:
    image: mongo:latest
    ports: [27017]
  rabbitmq:
    image: rabbitmq:3-management
    ports: [5672, 15672]
  qdrant:
    image: qdrant/qdrant:latest
    ports: [6333, 6334]
  redis:
    image: redis:latest
    ports: [6379]
  temporal:
    image: temporalio/temporal:latest
    ports: [7233, 8233]
    args: "server start-dev --ip 0.0.0.0 --ui-port 8233"

services:
  backend:
    directory: backend
    port: 8005
    type: python
    needs: [mongo]
    env_file: .env
    launch: "{python} -m run.dev"
    health: /api/health
    venv:
      install: "pip install -r requirements.txt"
      extras: ["../global_utils"]

  rag:
    directory: rag
    port: 13457
    type: python
    needs: [mongo, rabbitmq, qdrant]
    env_file: .env
    launch: "{python} -m bootstrap.flask_app"
    health: /api/health
    venv:
      install: "pip install -r requirements.txt"
      extras: ["../global_utils"]

  multi-agent:
    directory: multi-agent
    port: 8002
    type: python
    needs: [mongo, redis, temporal]
    env_vars: { HOSTNAME: "0.0.0.0" }
    launch: "mas api dev"
    health: /api/health
    venv:
      install: "pip install -e '.[all]'"
      extras: ["../global_utils"]

  sso:
    directory: shared-resources/sso-backend
    port: 13456
    type: python
    needs: []
    env_file: .env
    launch: "{python} app.py"
    health: /api/health
    venv:
      install: "pip install -r requirements.txt"
      extras: ["../../global_utils"]

  ui:
    directory: ui
    port: 5000
    type: node
    needs: []
    env_file: .env.local
    launch: "npm start"

  celery-worker:
    directory: rag
    type: python
    shares_venv_with: rag
    needs: [mongo, rabbitmq, qdrant]
    env_file: .env
    launch: >-
      celery -A infrastructure.celery.app worker
      -Q document_queue,slack_queue -l info

  temporal-worker:
    directory: multi-agent
    type: python
    shares_venv_with: multi-agent
    needs: [mongo, redis, temporal]
    launch: "mas temporal-worker --threads 20"

groups:
  all:         [backend, rag, multi-agent, sso, ui, celery-worker, temporal-worker]
  services:    [backend, rag, multi-agent, sso, ui]
  workers:     [celery-worker, temporal-worker]
  agents:      [multi-agent, temporal-worker]
  rag-stack:   [rag, celery-worker]
  backend-only: [backend, sso]

proxy_routes:
  /api1: { service: rag,          rewrite: /api }
  /api2: { service: multi-agent,  rewrite: /api }
  /api3: { service: sso,          rewrite: /api }
  /api4: { service: backend,      rewrite: /api }

env_templates:
  rag/.env:
    hostname_local: "127.0.0.1"
    port: "13457"
  shared-resources/sso-backend/.env:
    keycloak_base_url: "https://auth.stage.redhat.com/auth"
    client_id: "<REPLACE_WITH_YOUR_CLIENT_ID>"
    client_secret: "<REPLACE_WITH_YOUR_CLIENT_SECRET>"
    keycloak_realm: "EmployeeIDP"
    hostname_local: "127.0.0.1"
    port: "13456"
    frontend_url: "http://127.0.0.1:5000"
    backend_env: "development"
  ui/.env.local:
    DEV_PORT: "5000"
    DEV_HOST: "0.0.0.0"
    RAG_HOST: "http://127.0.0.1:13457"
    MULTIAGENT_HOST: "http://127.0.0.1:8002"
    SSO_HOST: "http://127.0.0.1:13456"
    BACKEND_HOST: "http://127.0.0.1:8005"
```

**Adding a new service = adding one YAML block. Zero code changes.**

---

### 3.3 Hexagonal Tooling Architecture

**Proposed file structure:**

```
local-development/
├── services.yaml                    # Single source of truth (data)
├── unifai-dev                       # Entry point (thin CLI)
├── LOCAL_DEV_GUIDE.md               # Documentation
│
├── devtool/                         # Python package
│   ├── __init__.py
│   │
│   ├── domain/                      # Core (no external deps)
│   │   ├── models.py                # Service, InfraComponent, ServiceGroup
│   │   └── registry.py              # Loads services.yaml → domain models
│   │
│   ├── ports/                       # Interfaces (ABCs)
│   │   ├── container_runtime.py     # start/stop/status containers
│   │   ├── session_manager.py       # create sessions/windows/panes
│   │   ├── process_manager.py       # launch/kill processes
│   │   └── venv_manager.py          # create/verify venvs
│   │
│   ├── adapters/                    # Implementations (swappable)
│   │   ├── podman.py                # ContainerRuntime → Podman
│   │   ├── docker.py                # ContainerRuntime → Docker
│   │   ├── tmux.py                  # SessionManager → tmux
│   │   ├── foreground.py            # SessionManager → plain terminal
│   │   └── venv.py                  # VenvManager → python -m venv
│   │
│   ├── services/                    # Application services
│   │   ├── orchestrator.py          # Composes ports for start/stop flows
│   │   ├── health_checker.py        # Probes service health endpoints
│   │   ├── recovery.py              # Dependency-aware restart engine
│   │   ├── env_generator.py         # Generates .env from registry
│   │   └── python_detector.py       # Finds Python 3.11-3.13
│   │
│   └── cli.py                       # Driving adapter: parses args, calls orchestrator
│
└── tests/
    ├── test_registry.py
    ├── test_orchestrator.py
    └── test_env_generator.py
```

**Hexagonal diagram:**

```
                    ┌──────────────────────────────────────────────┐
                    │              CLI  (cli.py)                    │
                    │         DRIVING ADAPTER                       │
                    └─────────────────┬────────────────────────────┘
                                      │
                                      ▼
  ┌────────────────────────────────────────────────────────────────────────┐
  │                                                                        │
  │                    APPLICATION  (orchestrator.py)                       │
  │                                                                        │
  │    start(services, windows, foreground)                                │
  │    stop(session)                                                       │
  │    restart(service)  ← dependency-aware                               │
  │    status() → dict   ← health dashboard                              │
  │    doctor() → report ← full diagnostic                               │
  │                                                                        │
  │    Uses ONLY ports (ABCs), never concrete adapters                    │
  │                                                                        │
  └───────┬──────────┬────────────┬───────────────┬────────────────────────┘
          │          │            │               │
          ▼          ▼            ▼               ▼
  ┌────────────┐┌──────────┐┌──────────────┐┌──────────────┐
  │ Container  ││ Session  ││   Process    ││    Venv      │
  │ Runtime    ││ Manager  ││   Manager    ││   Manager    │
  │ (PORT)     ││ (PORT)   ││  (PORT)      ││  (PORT)      │
  └─────┬──┬──┘└────┬──┬──┘└──────────────┘└──────────────┘
        │  │        │  │
        ▼  ▼        ▼  ▼            DRIVEN ADAPTERS
  ┌──────┐┌──────┐┌──────┐┌────────────┐
  │Podman││Docker││ tmux ││ Foreground │
  └──────┘└──────┘└──────┘└────────────┘

                    ┌──────────────────┐
                    │  services.yaml   │
                    │  (pure data)     │
                    └──────────────────┘
```

Why this is better:
- **Testable**: Mock the ports, test the orchestrator logic without Docker/tmux
- **Swappable**: Replace Podman with Docker, tmux with Zellij, without touching orchestration
- **Readable**: Each file is < 100 lines with one clear responsibility
- **Python-native**: Your team is a Python team — the tooling should match

---

### 3.4 Composable tmux Interface

**Current: binary choice**
```
--service backend         →  ONE service, foreground, no tmux
(no flags)                →  ALL 7 services, fixed hardcoded layout
```

**Proposed: full composability**
```bash
# Pick any combination of services
unifai-dev start backend rag

# Use predefined groups
unifai-dev start --group workers
unifai-dev start --group agents --group rag-stack

# Custom window layout
unifai-dev start \
    --window "api=backend,rag,sso" \
    --window "agents=multi-agent,temporal-worker" \
    --window "frontend=ui" \
    --window "workers=celery-worker"

# Single service foreground (replaces --service)
unifai-dev start backend --fg
```

**Example: `unifai-dev start backend rag --window "workers=celery-worker"`**

```
┌───────────────────────────────────────────────┐
│  tmux session: unifai-dev                     │
│                                               │
│  Window 0: "services"                         │
│  ┌─────────────────┬─────────────────┐        │
│  │  backend :8005  │  rag :13457     │        │
│  └─────────────────┴─────────────────┘        │
│                                               │
│  Window 1: "workers"                          │
│  ┌───────────────────────────────────┐        │
│  │  celery-worker                    │        │
│  └───────────────────────────────────┘        │
└───────────────────────────────────────────────┘
```

The tmux adapter reads the service registry, gets the launch command, directory, env file, and infra dependencies, builds panes dynamically, and auto-tiles with `select-layout tiled`. Zero hardcoded geometry.

---

### 3.5 Recovery and Observability

The current tooling covers **setup** and **start** but nothing after that. The steady-state phase (where developers spend 99% of their time) has zero support.

**`unifai-dev status` — health dashboard:**

```
$ unifai-dev status

  INFRASTRUCTURE
  ✔ MongoDB       :27017    running   (up 2h 15m)
  ✔ RabbitMQ      :5672     running   (up 2h 15m)
  ✖ Redis         :6379     STOPPED

  SERVICES
  ✔ backend       :8005     healthy   (v1.0.0)  tmux:0.4
  ✔ rag           :13457    healthy   (v1.0.0)  tmux:0.0
  ✖ multi-agent   :8002     DOWN                tmux:0.2

  ISSUES
  1. Redis stopped → multi-agent + temporal-worker affected
     Fix: unifai-dev infra start redis
  2. multi-agent not responding
     Fix: unifai-dev restart multi-agent
```

This is possible because every service already exposes a health endpoint:
- `backend`: `GET /` → `{"status": "ok"}`
- `rag`: `GET /` → `{"status": "ok"}`
- `multi-agent`: `GET /` → `{"status": "ok"}`
- `sso`: `GET /` → `{"status": "ok"}`

**`unifai-dev restart <service>` — dependency-aware restart:**

```
$ unifai-dev restart multi-agent

  ⏹  Stopping multi-agent (tmux pane 0.2)...
  🔄 Checking dependencies...
     ✔ MongoDB  — running
     ✖ Redis    — STOPPED → restarting...
     ✔ Redis    — running
     ✔ Temporal — running
  🚀 Restarting multi-agent...
  ⏳ Health check...
  ✔ multi-agent healthy
```

The restart engine uses the dependency graph from `services.yaml`:

```
RECOVERY ORDER:
───────────────
  Layer 0 — Infrastructure    Fix containers first
  Layer 1 — Services          Restart services that depend on fixed infra
  Layer 2 — Workers           Restart workers last
```

**`unifai-dev restart --failed` — fix everything broken:**

```
$ unifai-dev restart --failed

  Scanning... found 2 issues
  ✔ Redis container restarted
  ✔ multi-agent restarted in tmux pane 0.2
  ✔ All services healthy
```

**`unifai-dev logs <service>` — view logs without tmux navigation:**

```
$ unifai-dev logs multi-agent
# Uses: tmux capture-pane -t <pane_id> -p -S -100

$ unifai-dev logs multi-agent --follow
# Uses: tmux pipe-pane
```

**`unifai-dev doctor` — full diagnostic:**

```
$ unifai-dev doctor

  SYSTEM
  ✔ Python 3.12.4    ✔ Node 22.1.0    ✔ tmux 3.4    ✔ Podman 5.0.1

  VENVS
  ✔ backend/venv       3.12.4   ✔ global_utils
  ✔ rag/venv           3.12.4   ✔ global_utils
  ✔ multi-agent/venv   3.12.4   ✔ global_utils
  ⚠ sso-backend/.env   client_id is still a placeholder!

  INFRASTRUCTURE
  ✔ MongoDB :27017     ✔ RabbitMQ :5672    ✖ Redis :6379 stopped

  RECOMMENDATIONS
  1. Fix Redis:      unifai-dev infra start redis
  2. Fix SSO creds:  edit shared-resources/sso-backend/.env
```

---

### 3.6 Proposed CLI

```
unifai-dev
├── start [services...]              # Start services (tmux or --fg)
│   ├── --fg                         # Foreground single service
│   ├── --group <name>               # Use predefined group from registry
│   ├── --window "name=svc1,svc2"    # Custom tmux window layout
│   └── --setup-venv                 # Create venvs first
│
├── stop                             # Stop tmux session
│
├── restart <service>                # Dependency-aware single restart
│   └── --failed                     # Auto-fix all broken services
│
├── status                           # Health dashboard
├── logs <service> [--follow]        # View/tail service logs
├── doctor                           # Full diagnostic
│
├── infra
│   ├── start [containers...]        # Start infrastructure
│   │   └── --for <service>          # Only what a service needs
│   ├── stop                         # Stop all containers
│   └── status                       # Container status
│
├── venv
│   ├── setup [service]              # Create venv(s)
│   └── check                        # Verify Python versions match
│
├── env
│   ├── generate [--force]           # Create/regenerate .env files
│   └── show <service>               # Print env config
│
└── destroy                          # Kill everything
```

---

### 3.7 What "Adding a New Service" Looks Like

**Current (7+ files, 15+ edits):**

| # | File | Edit |
|---|------|------|
| 1 | `start-unifai-dev.sh` | Add `case` block in single-service mode (~line 242) |
| 2 | `start-unifai-dev.sh` | Add port to `PORTS` string (~line 424) |
| 3 | `start-unifai-dev.sh` | Add `verify_venv` call (~line 406) |
| 4 | `start-unifai-dev.sh` | Add tmux pane creation (~line 446+) |
| 5 | `start-unifai-dev.sh` | Add `ensure_container` calls (~line 411+) |
| 6 | `start-infra.sh` | Add to `SERVICE_INFRA` mapping (~line 68) |
| 7 | `apply-local-dev-changes.py` | Add venv table entry (~line 163) |
| 8 | `apply-local-dev-changes.py` | Add `.env` generation (~line 265+) |
| 9 | `apply-local-dev-changes.py` | Add source patch if needed |
| 10 | `LOCAL_DEV_GUIDE.md` | Update 5+ tables |
| 11 | `vite.config.ts` | Add proxy route if it has an API |

**Proposed (1 file, 1 block):**

```yaml
# services.yaml — just add this block:
analytics:
  directory: analytics
  port: 8006
  type: python
  needs: [mongo, redis]
  env_file: .env
  launch: "{python} -m run.dev"
  health: /api/health
  venv:
    install: "pip install -r requirements.txt"
    extras: ["../global_utils"]
```

Done. The orchestrator reads the registry and handles venv, infra, tmux, port cleanup, health checks, and recovery automatically.

---

## 4. Priority Matrix

### Must-fix before merge (high risk)

| # | Issue | Why |
|---|-------|-----|
| 1 | Source patching must be manually reverted | One forgotten `git checkout` → broken production. Add `bind_host` to `SharedConfig` and eliminate patching entirely. |
| 2 | Committed `vite.config.ts` broken without `.env.local` | `process.env.RAG_HOST` resolves to `undefined` without tooling. Add fallback URLs. |
| 3 | Commit `/api3` proxy route | Don't inject it at runtime with fragile regex. Commit it alongside `/api1`, `/api2`, `/api4`. |

### Should-fix (high impact, moderate effort)

| # | Issue | Why |
|---|-------|-----|
| 4 | Deduplicate `ensure_container` | Full-stack mode should call `start-infra.sh` instead of reimplementing container management. |
| 5 | Validate that source patches actually matched | `str.replace` silently does nothing if the target string changed. At minimum, assert the replacement was applied. |
| 6 | Remove `ALL_VENV_SERVICES` dead code | Line 160 of `apply-local-dev-changes.py` — unused variable. |

### Consider for follow-up (architectural improvements)

| # | Improvement | Impact |
|---|-------------|--------|
| 7 | Service registry (`services.yaml`) | Eliminates shotgun surgery. One file to add a service. |
| 8 | Composable tmux (`start svc1 svc2 --window ...`) | Full developer control over layout. |
| 9 | Recovery commands (`status`, `restart`, `doctor`) | Covers the 99% of time developers spend after startup. |
| 10 | Python CLI rewrite | Testable, typed, modular — matches the team's language. |
| 11 | Python detection deduplication | One implementation shared by bash entry point and Python scripts. |

---

*This review is constructive and not a blocker on the overall direction — the tooling solves a real need and the documentation is thorough. The suggestions aim to make the system safer (no source patching), cleaner (single source of truth), and more flexible (composable tmux + recovery) as the team scales.*

yhabushi79 (dismissed):
tested several times with Shani.
works fine.
i guess we need to let new people use it for more confidance.

local-development/devtool/adapters/container_base.py
```diff
@@ -0,0 +1,254 @@
...
+        result = subprocess.run([*cmd, "info"], capture_output=True)
+        if result.returncode == 0:
+            return SubprocessContainerRuntime(cmd)
+        raise RuntimeError(
+            f"UNIFAI_CONTAINER_RUNTIME is set to '{env_override}' "
+            f"but '{env_override} info' failed. "
+            f"Verify the command works in your terminal."
+        )
+
+    from devtool.adapters.podman import PodmanRuntime
```
sfiresht: isn't it more efficient to put the imports as lazy imports after the if?

local-development/devtool/adapters/container_base.py
```diff
@@ -0,0 +1,254 @@
...
+                ["podman", "machine", "start"],
+                capture_output=True,
+            )
+            check = subprocess.run(
+                ["podman", "info"],
+                capture_output=True,
+            )
+            if check.returncode == 0:
+                return PodmanRuntime()
+
```
sfiresht: I see that in podman we do more checks, ay reason we invest more in the podman then in docker? is there anything mssing from docker?

sfiresht (dismissed):
though I've added comments, I think they're not that crucial, approving (you'll still need to resolve these comments though)

local-development/devtool/adapters/container_base.py
shanitzvii: The extrachecks for podman handles the podman machine concept, which is specific to mac OS and windows, and not requires privileges. it's for the usecase where the VM that runs podman containers exist but was stopped. This doesn't apply on Docker.

local-development/devtool/adapters/container_base.py
shanitzvii: fixed. if you can re approve :) @sfiresht

sfiresht (dismissed):

sfiresht:
in my VM it failed and we had to select the image from docker.io in order for it to work.
in addition please add local-development/.dev-secret-key to the git ignore.

sfiresht (dismissed):

sfiresht (dismissed):

yhabushi79 (dismissed):

local-development/devtool/services/python_detector.py
odaiodeh, line 1: @shanitzvii is python detector should be in services? is a domain code or an adapter code?

odaiodeh:
@shanitzvii 

# PR Review: `GENIE-1501_local_dev_guide`

## Overview

This PR introduces a CLI-based local development tool (`unifai-dev`) following hexagonal architecture, along with a dev-user authentication bypass for local development. The overall architecture is solid, but there are issues with domain purity, SOLID violations, missing enums, security concerns, code cleanliness, and cross-platform readiness that should be addressed before merge.

**Files reviewed:** 50+ files across `local-development/`, `shared-resources/identity/`, and `ui/`.

---

## Table of Contents

- [1. Hexagonal Architecture Violations](#1-hexagonal-architecture-violations)
- [2. SOLID Violations](#2-solid-violations)
- [3. Missing Enums and Magic Strings](#3-missing-enums-and-magic-strings)
- [4. Pydantic vs Dataclasses](#4-pydantic-vs-dataclasses)
- [5. Security Concerns](#5-security-concerns)
- [6. Error Handling Inconsistencies](#6-error-handling-inconsistencies)
- [7. Code Cleanliness](#7-code-cleanliness)
- [8. Missing Type Annotations](#8-missing-type-annotations)
- [9. Code Duplication](#9-code-duplication)
- [10. Edge Cases and Robustness](#10-edge-cases-and-robustness)
- [11. Shared-Resources and UI Changes](#11-shared-resources-and-ui-changes)
- [12. Cross-Platform Compatibility](#12-cross-platform-compatibility)
- [13. What's Done Well](#13-whats-done-well)
- [Summary Table](#summary-table)

---

## 1. Hexagonal Architecture Violations

### 1.1 `Registry` lives in `domain/` but performs I/O and depends on external libraries

**File:** `devtool/domain/registry.py`

The Registry is placed in the domain layer but it:
- **Imports `yaml`** (external library) — lines 9-16
- **Reads files from disk** (`open(yaml_path)`) — line 34
- **Reads environment variables** (`os.environ.get`) — lines 37, 156-157
- **Prints to `stderr`** and calls `SystemExit` — lines 12-16

In hexagonal architecture, the domain layer must be pure: no I/O, no external dependencies, no infrastructure concerns.

**Recommendation:** Split into:
- A pure domain `Registry` class that takes already-parsed data (in-memory lookups only)
- A `YamlRegistryLoader` adapter/factory in the services or adapters layer that handles file I/O, YAML parsing, and env var resolution

### 1.2 `Orchestrator` imports service modules directly instead of through ports

**File:** `devtool/services/orchestrator.py`, lines 17-19

```python
from devtool.services import env_generator, python_detector
from devtool.services.constants import SESSION_NAME
from devtool.services.shell_utils import resolve_bash
```

Additionally, there are 5 inline imports inside methods (lines 168, 277, 412, 476, 894) that make the dependency graph invisible. Services like `env_generator`, `python_detector`, `health_checker`, and `Recovery` are used directly rather than injected through ports.

**Recommendation:** Either define ports for these capabilities or inject them as constructor dependencies. At minimum, move all imports to the top of the file.

---

## 2. SOLID Violations

### 2.1 SRP — `Orchestrator` is a God class (~974 lines, 100+ print statements)

**File:** `devtool/services/orchestrator.py`

This single class handles: start, stop, attach, shell, exec, infra management (start/stop/reset/status/logs), venv management (setup/sync/check), env generation (generate/show), init (6-step wizard), clean, restart, status, doctor, port checking, summary printing, and shell completion installation.

**Recommendation:** Extract into focused services:
- `StartupService` — start flow orchestration
- `InfraService` — infra start/stop/reset/status/logs
- `VenvService` — setup/sync/check
- `EnvService` — generate/show/resolve
- `DiagnosticService` — status/doctor
- Keep `Orchestrator` as a thin facade

### 2.2 SRP — `env_generator.py` mixes too many concerns

**File:** `devtool/services/env_generator.py`

This single module handles: file generation, placeholder checking, missing key checking, auto-generate resolution, file rewriting, shared secret management, local-auth alignment, and file display — at least 4 different responsibilities.

### 2.3 OCP — `detect_runtime()` is not open for extension

**File:** `devtool/adapters/container_base.py`, lines 195-255

Adding a new runtime (e.g., `nerdctl`) requires modifying this function. A runtime-detector registry or strategy pattern would be more extensible.

### 2.4 DIP — `Recovery` directly imports service modules

**File:** `devtool/services/recovery.py`, line 12

```python
from devtool.services import health_checker
```

The `Recovery` class depends on a concrete service module rather than an abstraction.

---

## 3. Missing Enums and Magic Strings

### 3.1 `ServiceHealth.status` is `str`, not an enum

**File:** `devtool/domain/models.py`, line 76

```python
status: str  # uses "healthy", "DOWN", "unhealthy", "no port"
```

These magic strings are then compared throughout `health_checker.py` (lines 60, 69, 80, 172-182) and `recovery.py` (line 64). A typo in any of these strings causes a silent bug.

**Recommendation:** Create a `ServiceStatus` enum:
```python
class ServiceStatus(Enum):
    HEALTHY = "healthy"
    UNHEALTHY = "unhealthy"
    DOWN = "down"
    NO_PORT = "no_port"
```

### 3.2 `"identity"` hardcoded as a magic string (3 occurrences)

**File:** `devtool/services/env_generator.py`, lines 36, 81, 158

```python
is_identity_local = local_auth and service.name == "identity"
```

**Recommendation:** Extract to a constant or make local-auth a per-service config flag in `services.yaml`.

### 3.3 `_KEYCLOAK_KEYS` is a fragile plain set of strings

**File:** `devtool/services/env_generator.py`, line 26

```python
_KEYCLOAK_KEYS = {"keycloak_base_url", "client_id", "client_secret", "keycloak_realm"}
```

Adding a new Keycloak config key requires remembering to update this set.

---

## 4. Pydantic vs Dataclasses

### All domain models use `dataclasses` without validation

**File:** `devtool/domain/models.py`

While `dataclass(frozen=True)` provides immutability, Pydantic would provide:
- **Validation on construction** (e.g., ensure `port > 0`, `image` is non-empty)
- **Type coercion** (YAML values currently pass through with no validation)
- **`model_validate()`** to replace the 30+ line manual `_parse_services` / `_parse_infra` methods in `registry.py`
- **Serialization** for free

Currently, `Registry._parse_services()` (lines 180-202) manually extracts dict fields with `.get()` and no validation — bad YAML values silently produce invalid domain objects.

---

## 5. Security Concerns

### 5.1 Shell injection risk in `exec_in_context`

**File:** `devtool/services/orchestrator.py`, lines 199-207

```python
user_cmd = " ".join(command)
shell_cmd = f"{context} && {user_cmd}"
bash = resolve_bash()
os.execvp(bash, [bash, "-c", shell_cmd])
```

User CLI arguments are `" ".join`'d into a `bash -c` string. Arguments containing spaces, quotes, semicolons, or shell metacharacters will be misinterpreted. `shlex.quote()` is never used anywhere in the codebase (confirmed: 0 occurrences).

### 5.2 Unquoted paths in shell commands

**File:** `devtool/services/orchestrator.py`, lines 808-820

```python
parts.append(f"cd {svc_dir}")
```

Paths with spaces or special characters will break. Same issue in `tmux.py` line 178:

```python
wrapped = f"{cmd} 2>&1 | tee {log_path}"
```

**Recommendation:** Use `shlex.quote()` for all paths and commands interpolated into shell strings.

### 5.3 Shared secret file written with default permissions

**File:** `devtool/services/env_generator.py`, lines 121-123

```python
key = secrets.token_hex(32)
secret_path.write_text(key + "\n")
```

On multi-user systems, the default umask (often `022`) makes this file world-readable. Should use `os.open()` with `0o600` permissions or `secret_path.chmod(0o600)` after creation.

### 5.4 `component.command.split()` doesn't handle quoting

**File:** `devtool/adapters/container_base.py`, lines 47-48

```python
if component.command:
    cmd.extend(component.command.split())
```

The Temporal command in `services.yaml` is `"server start-dev --ip 0.0.0.0 --ui-port 8233"` — works today, but any command with quoted arguments containing spaces would break. Should use `shlex.split()`.

---

## 6. Error Handling Inconsistencies

The codebase mixes **4 different error-handling patterns** with no clear convention:

| Pattern | Where Used | Example |
|---------|-----------|---------|
| `raise SystemExit(msg)` | `_validate_start`, `venv_check`, `init`, `registry.py` | Aborts the entire CLI |
| `raise RuntimeError(msg)` | Adapters, `python_detector` | Propagates up |
| `print(error) + return` | `attach`, `logs`, `restart`, `stop` | Silent success exit code 0 |
| `print(error) + continue` | `start` venv setup, `env_generate` | Best-effort continuation |

### Specific issue: `doctor` calls `venv_check()` which raises `SystemExit`

**File:** `devtool/services/orchestrator.py`, lines 437-438

```python
print("\nVirtual environments:")
self.venv_check()   # <-- raises SystemExit(1) if any venv fails
```

This **aborts the diagnostic** mid-run, skipping env file and port checks. `doctor` should complete all checks and report a summary, not abort on the first venv error.

### `_detect_python` has unhandled exceptions

**File:** `devtool/services/orchestrator.py`, lines 775-777

```python
ver_out = subprocess.check_output([python, "--version"], text=True).strip()
version = ver_out.split()[-1]
minor = ".".join(version.split(".")[:2])
```

If `check_output` fails or produces unexpected output, this raises `CalledProcessError` or `IndexError` with no user-friendly message — despite `find_python` already validating the binary earlier.

### `_resolve_root` doesn't validate `UNIFAI_ROOT`

**File:** `devtool/cli.py`, lines 114-117

```python
alt = os.environ.get("UNIFAI_ROOT", "").strip()
if alt:
    return Path(alt).expanduser().resolve()
```

When `UNIFAI_ROOT` is set, the `rag`/`ui` directory checks are skipped entirely. An invalid `UNIFAI_ROOT` silently produces a bad root path.

### Unknown service names produce raw `KeyError` tracebacks

**File:** `devtool/domain/registry.py`, lines 52-58

The `get_service` method raises `KeyError` with a nice message, but Typer doesn't catch `KeyError` — the user sees a full Python traceback instead of a clean error.

---

## 7. Code Cleanliness

### 7.1 Resource leak with `noqa` suppression

**File:** `devtool/adapters/container_base.py`, line 160

```python
stderr_dest = open(self._log_file, "a")  # noqa: SIM115
```

The `noqa` acknowledges the linter warning but doesn't fix the underlying issue. Use a context manager.

### 7.2 Copy-paste comments in `vite.config.ts`

**File:** `ui/vite.config.ts`, lines 49-61

```typescript
// Proxy for api2 (assuming this is still local or another service)  ← wrong
'/api3': {
    target: process.env.IDENTITY_HOST,//'http://127.0.0.1:13457', // Your second backend  ← wrong
    rewrite: (path) => path.replace(/^\/api3/, '/api'), // This rewrites /api2 to nothing  ← wrong
},
'/api4': {
    target: process.env.BACKEND_HOST,//'http://127.0.0.1:13457', // Your second backend  ← wrong
    rewrite: (path) => path.replace(/^\/api4/, '/api'), // This rewrites /api2 to nothing  ← wrong
},
```

Every comment is a stale copy-paste from the `/api2` block. Inline commented-out URLs are also misleading.

### 7.3 PEP 8 violation in `auth_manager.py`

**File:** `shared-resources/identity/utils/auth_manager.py`, line 8

```python
import os, uuid, logging
```

Multiple imports on one line violates PEP 8.

### 7.4 f-strings in logger calls

**File:** `shared-resources/identity/utils/auth_manager.py` (throughout, e.g., lines 79, 185, 216, 225, 228, 327, 376)

```python
logger.error(f"Failed to setup Keycloak client: {e}")
```

f-strings evaluate even when the log level is disabled. Use lazy formatting: `logger.error("Failed to setup Keycloak client: %s", e)`.

### 7.5 Verbose logic that can be simplified

**File:** `shared-resources/identity/utils/auth_manager.py`, lines 380-403

```python
def _check_admin_permission(self, user: dict) -> bool:
    if not user:
        return False
    allowed_users = config.get('admin_allowed_users', [])
    if not allowed_users:
        return False
    username = user.get('username') or user.get('sub')
    if username and username in allowed_users:
        return True
    return False
```

Can be reduced to:
```python
def _check_admin_permission(self, user: dict) -> bool:
    if not user:
        return False
    allowed = config.get('admin_allowed_users', [])
    identifier = user.get('username') or user.get('sub')
    return bool(identifier and identifier in allowed)
```

### 7.6 `get_user_info` builds dict key-by-key

**File:** `shared-resources/identity/utils/auth_manager.py`, lines 301-314

Manually copies 7 keys one at a time from `session_data` to `user`. Should use a Pydantic model, a `TypedDict`, or at least a dict comprehension:
```python
_USER_FIELDS = ('username', 'email', 'name', 'sub', 'session_created_at', 'session_expires_at', 'token_expires_at')
return {k: session_data.get(k) for k in _USER_FIELDS}
```

### 7.7 Magic number 20 in `check_unresolved`

**File:** `devtool/services/env_generator.py`, line 217

```python
value_start = stripped[eq + 1:eq + 20]
```

Unexplained magic number. Why only check the first 19 characters of the value?

### 7.8 `recovery.py` has unused `root: Path` parameter

**File:** `devtool/services/recovery.py`, lines 31, 57

Both `restart_service` and `restart_failed` accept `root: Path` but never use it.

---

## 8. Missing Type Annotations

| File | Location | Issue |
|------|----------|-------|
| `health_checker.py` | line 92 | `component` param has no type hint |
| `orchestrator.py` | line 933 | `infra: list` — bare `list` without type parameter |
| `dev_oauth_client.py` | all methods | No return types, no parameter types |
| `auth_manager.py` | lines 22, 31, 55, etc. | Methods lack return type annotations |
| `cli.py` | line 127 | `_load_registry` return type missing |

---

## 9. Code Duplication

### 9.1 Venv error-accumulation pattern repeated 4 times

**File:** `devtool/services/orchestrator.py`

The same try/except/append-errors/print-summary pattern appears in:
- `start()` — lines 86-97
- `venv_setup()` — lines 305-328
- `venv_sync()` — lines 342-356
- `venv_check()` — lines 358-371

**Recommendation:** Extract a helper like `_run_for_services(action, services)` that handles the try/except/accumulate/report pattern.

### 9.2 `_create_orchestrator` vs `_load_registry` in CLI

**File:** `devtool/cli.py`, lines 127-161

Both functions create a `Registry()`. `_load_registry` creates just the Registry for read-only commands; `_create_orchestrator` creates Registry + all adapters. The Registry creation is duplicated and could share a common path.

### 9.3 Strategy branching in venv adapter

**File:** `devtool/adapters/venv.py`

The `create()` and `_install_deps()` methods both switch on `VenvStrategy` with similar branching logic (lines 26-46, 129-152).

---

## 10. Edge Cases and Robustness

### 10.1 `venv.py` `verify()` can `IndexError` on unexpected Python output

**File:** `devtool/adapters/venv.py`, lines 67-68

```python
venv_ver = result.stdout.strip().split()[-1]
venv_minor = ".".join(venv_ver.split(".")[:2])
```

If `result.stdout` is empty or malformed, `split()[-1]` raises `IndexError`.

### 10.2 `logs()` loads entire log file into memory

**File:** `devtool/services/orchestrator.py`, line 407

```python
print(log_path.read_text(), end="")
```

For large log files, this could use excessive memory. Consider streaming with a subprocess call to `cat` or reading in chunks.

### 10.3 `restart_service` pane matching via substring is fragile

**File:** `devtool/adapters/tmux.py`, line 146

```python
if service_name in capture.stdout:
```

This can false-positive if a service name like `"rag"` appears in unrelated tmux scrollback text.

### 10.4 `process.py` may signal duplicate PIDs

**File:** `devtool/adapters/process.py`, lines 67-78

`kill_processes` iterates `pids` without deduplication — if `lsof` returns the same PID multiple times, it sends SIGTERM/SIGKILL twice.

### 10.5 `_custom` venv commands split naively

**File:** `devtool/adapters/venv.py`, lines 158-160

```python
for cmd_template in service.venv.commands:
    cmd_str = cmd_template.replace("{python}", python)
    self._run(cmd_str.split(), svc_dir, log_file)
```

`str.split()` doesn't handle quoted arguments. A YAML command like `"pip install 'package with spaces'"` would break. Should use `shlex.split()`.

### 10.6 Mutable class-level default in `app_config.py`

**File:** `shared-resources/identity/config/app_config.py`, line 21

```python
admin_allowed_users: list = []
```

Bare `list = []` as a class attribute is a mutable default shared across instances.

---

## 11. Shared-Resources and UI Changes

### 11.1 `DevOAuthClient` lacks error handling and type safety

**File:** `shared-resources/identity/utils/dev_oauth_client.py`

- No error handling at all (if something fails, raw exceptions propagate)
- No type annotations on any method
- Returns plain dicts instead of typed models
- No base class or protocol

### 11.2 Login.tsx — no error state for config fetch failure

**File:** `ui/client/src/pages/Login.tsx`, lines 13-17

```tsx
useEffect(() => {
    api.get('/auth/config')
      .then(res => setLocalAuth(res.data.local_auth))
      .catch(() => setLocalAuth(false));
}, []);
```

On network failure, silently defaults to `false` (SSO mode). The user sees no indication that the config fetch failed. Consider showing a retry button or error message.

### 11.3 `logout` route tries Keycloak revocation even in local-auth mode

**File:** `shared-resources/identity/utils/auth_manager.py`, lines 192-225

When `local_auth_enabled=true`, the logout route still attempts to call `config.keycloak_base_url` (which is empty) for token revocation. The `try/except` catches it, but it's a wasted HTTP call and a noisy warning log.

**Recommendation:** Guard the Keycloak revocation block with `if not config.local_auth_enabled:`.

---

## 12. Cross-Platform Compatibility

The tool currently assumes a Unix/macOS environment. Several areas need attention for Windows and broader Linux support.

### 12.1 Hard Unix dependencies — no Windows fallback

| Dependency | Where Used | Windows Issue |
|-----------|-----------|---------------|
| `tmux` | `adapters/tmux.py` (19 subprocess calls) | Does not exist on Windows |
| `bash` | `shell_utils.py`, `orchestrator.py`, `foreground.py` | Not available by default on Windows |
| `os.execvp` | `orchestrator.py` lines 197, 207; `foreground.py` line 41 | Not supported on Windows |
| `signal.SIGTERM` / `signal.SIGKILL` | `process.py` lines 69, 76 | `SIGKILL` does not exist on Windows |
| `lsof` | `process.py` line 23 | Unix-only |
| `ss` | `process.py` line 38 | Linux-only |
| `/proc/net/tcp` | `process.py` lines 106-118 | Linux-only |
| `source venv/bin/activate` | `orchestrator.py` line 815 | Windows uses `venv\Scripts\activate.bat` |
| `venv/bin/pip`, `venv/bin/python` | `venv.py` lines 57, 143-149 | Windows uses `venv\Scripts\pip.exe` |
| `tee` | `tmux.py` line 178, `foreground.py` line 39 | Not available on Windows |

**Recommendations:**

1. **Abstract platform-specific operations into the port layer.** Create a `Platform` port/protocol:
   ```python
   class Platform(ABC):
       @abstractmethod
       def venv_python_path(self, venv_dir: Path) -> Path: ...
       @abstractmethod
       def venv_activate_command(self, venv_dir: Path) -> str: ...
       @abstractmethod
       def kill_process(self, pid: int, force: bool = False) -> None: ...
       @abstractmethod
       def find_port_occupants(self, port: int) -> list[PortOccupant]: ...
   ```

2. **Replace `os.execvp`** with `subprocess.run` (or `subprocess.Popen` + `sys.exit`) which works cross-platform. `os.execvp` replaces the process entirely and is not supported on Windows.

3. **Use `pathlib` consistently for venv paths:**
   ```python
   # Instead of hardcoding "venv/bin/python":
   if sys.platform == "win32":
       python_path = venv_dir / "Scripts" / "python.exe"
   else:
       python_path = venv_dir / "bin" / "python"
   ```
   Better: put this in the `Platform` adapter.

4. **Add a Windows `SessionManager` adapter.** Since tmux doesn't exist on Windows, provide an alternative:
   - `WindowsTerminalSessionManager` using Windows Terminal tabs/panes
   - `SubprocessSessionManager` that manages background processes directly
   - Or simply document that Windows users should use WSL2

5. **Replace `signal.SIGKILL` on Windows:**
   ```python
   if sys.platform == "win32":
       os.kill(pid, signal.SIGTERM)  # or use taskkill
   else:
       os.kill(pid, signal.SIGKILL)
   ```

6. **Port discovery on Windows:** Use `netstat -ano` or the `psutil` library instead of `lsof`/`ss`/`/proc`.

7. **Use `shlex.quote()` on Unix, avoid shell strings on Windows.** Windows command parsing is fundamentally different (`cmd.exe` vs bash). The cleanest approach is to avoid shell string composition entirely and use `subprocess.run(cmd_list)` wherever possible.

### 12.2 Path separator assumptions

**File:** `devtool/services/orchestrator.py`, line 812

```python
parts.append(f"cd {svc_dir}")
```

`Path` objects use `/` on Unix and `\` on Windows. Since this is composed into a shell string for `bash -c`, it only works on Unix. On Windows, you'd need `cmd /c cd ...` or PowerShell equivalents.

### 12.3 `.env` file sourcing is bash-specific

**File:** `devtool/services/orchestrator.py`, line 818

```python
parts.append(f"set -a && source {svc.env_file} 2>/dev/null; set +a")
```

`set -a`, `source`, and `2>/dev/null` are bash-specific. On Windows, environment loading requires a different mechanism (e.g., `python-dotenv` library, or PowerShell `Get-Content .env | ForEach-Object { ... }`).

**Recommendation:** Consider using `python-dotenv` to load `.env` files programmatically (cross-platform) instead of relying on shell `source` commands.

### 12.4 Recommended architecture for cross-platform support

```
ports/
    platform.py          # ABC for OS-specific operations
adapters/
    unix_platform.py     # lsof, signals, bash, venv/bin/
    windows_platform.py  # netstat, taskkill, cmd/powershell, venv\Scripts\
    wsl_platform.py      # delegate to unix inside WSL
```

The `_create_orchestrator()` function in `cli.py` should detect the platform and wire the appropriate adapter:

```python
import sys

def _create_platform() -> Platform:
    if sys.platform == "win32":
        from devtool.adapters.windows_platform import WindowsPlatform
        return WindowsPlatform()
    else:
        from devtool.adapters.unix_platform import UnixPlatform
        return UnixPlatform()
```

### 12.5 Minimal changes for immediate cross-platform progress

If full Windows support is not in scope right now, at minimum:

1. **Add `sys.platform` guards** with clear error messages when running on unsupported platforms
2. **Document** in the README/guide that Windows users need WSL2
3. **Replace `os.execvp`** with `subprocess.run` — this is a one-line fix that unblocks Windows
4. **Use `pathlib` for venv paths** instead of hardcoded `venv/bin/...` strings
5. **Consider `psutil`** as an optional dependency for cross-platform process management — it replaces `lsof`, `ss`, `/proc/net/tcp`, and `signal.SIGKILL` with a single cross-platform API

---

## 13. What's Done Well

- **Port interfaces** (`ports/`) are cleanly defined with clear ABC contracts and good docstrings
- **Enum usage** for `ServiceType`, `VenvStrategy`, `ContainerStatus`, `GenerateResult` is solid
- **Immutable domain models** with `frozen=True` dataclasses
- **Test coverage** is thorough with well-structured test classes, proper fixtures, and appropriate mocking at port boundaries
- **Tab completion** callbacks in the CLI are a great UX touch
- **Declarative `services.yaml`** is a clean config-driven approach
- **Graceful shutdown** with tmux `C-c` + timeout + force kill is well-implemented
- **Container runtime auto-detection** with env var override is practical
- **Health dashboard** with cross-referenced issue analysis (infra → affected services) is clever
- The **adapter pattern** for Docker/Podman/Tmux/Foreground is correct and extensible

shanitzvii:
summary of addressed issues from your review @odaiodeh 

all addressed except:
2.3 - we have a env variable for run time container.
3.3 - keycloak is only for identity and shouldnt change his config 
4 - preferred to keep as frozen data class - it adds complexity where we dont need - i dont expose those models as API  or serialize them to/from JSON. also they never validated against external/untrusted input since the data comes from  YAML config, already parsed.
section  7 - isnt in my scope, addresed 7.1 , 7.8 only.
9.2 - The "duplication" is a single line - YamlRegistryLoader.load(), while the _load_registry() is lightweight, no adapters, works without docker/tmux. only used by list and info commands- extracting a shared path would couple them and make the code harder to follow for negligible DRY benefit. 
10.6 - isnt my scope
12 - windows support will be in phase 2. applied the first 4 recommendations in 12.5.

local-development/devtool/adapters/container_base.py
```diff
@@ -0,0 +1,255 @@
...
+                    stderr=stderr_dest,
+                    check=True,
+                )
+            except subprocess.CalledProcessError:
+                log_hint = f" Check {self._log_file}" if self._log_file else ""
+                print(f"  ⚠ Command failed: {' '.join(cmd)}.{log_hint}")
+                raise
+
+
+def _format_duration(delta) -> str:
```
odaiodeh: this function can be in some util file in adapters that can help all other adapters too, since this is a gentic function, one more thing is the delta, its good practice to write the type of the delta, like delta: DateTimeDelta or what ever type is this.

local-development/devtool/adapters/container_base.py
```diff
@@ -0,0 +1,255 @@
...
+    hours = minutes // 60
+    remaining_min = minutes % 60
+    if hours < 24:
+        return f"{hours}h {remaining_min}m" if remaining_min else f"{hours}h"
+    days = hours // 24
+    remaining_hours = hours % 24
+    return f"{days}d {remaining_hours}h" if remaining_hours else f"{days}d"
+
+
+def detect_runtime() -> SubprocessContainerRuntime:
```
odaiodeh: Is this function outside the class? why?

shanitzvii: its a factory function - it create&return the class (SubprocessContainerRuntime) object, called by the cli.py.
wanted to keep the base class independent of runtime, without duplications. each container adapter inherit from the base class.

odaiodeh: you can make a directory called runtime or choose good name, there you you have the calsses, and also have a factory class that you call it to get the object

local-development/devtool/services/dotenv/inspector.py
```diff
@@ -0,0 +1,109 @@
+"""Read-only analysis of .env files: unresolved markers and missing keys."""
```
odaiodeh: why this not a class that gets the service? and has all these functions?

local-development/devtool/services/dotenv/generator.py
```diff
@@ -0,0 +1,96 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+
+from devtool.domain.registry import Registry
+
+from .common import ENV_HEADER, KEYCLOAK_KEYS, LOCAL_AUTH_SERVICE, GenerateResult
+from .inspector import check_missing_keys, check_unresolved
+from .local_auth import align_local_auth
+
```
odaiodeh: please make this file a class that gets the requieret service, and also show the type of service in the init

local-development/devtool/services/dotenv/display.py
```diff
@@ -0,0 +1,25 @@
+"""Display .env file contents to stdout."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from devtool.domain.models import Service
+
```
odaiodeh: please make this file a class that gets the requieret service, and also show the type of service in the init

local-development/devtool/services/dotenv/display.py
```diff
@@ -0,0 +1,25 @@
...
+
+def show(service: Service, root: Path) -> None:
```
odaiodeh: ok i see now this Service is something that represent the details of some service, i think you should change the name, like ServiceInfo or other name, since its a little bit confusing

local-development/devtool/services/dotenv/local_auth.py
```diff
@@ -0,0 +1,46 @@
+"""Identity-service local_auth alignment."""
```
odaiodeh: same comment as above

local-development/devtool/services/dotenv/resolver.py
```diff
@@ -0,0 +1,61 @@
+"""Mutation of existing .env files and shared secret management."""
```
odaiodeh: same comment as above

local-development/devtool/services/health_checker.py
```diff
@@ -0,0 +1,295 @@
...
+        status=st,
+        uptime=uptime,
+    )
+
+
+# ---------------------------------------------------------------------------
+# Full dashboard
+# ---------------------------------------------------------------------------
+
+def _build_dashboard(
```
odaiodeh: so all these factories needs to be done in more organzied way, as a class of factory, see above comments about how to do them

local-development/devtool/services/health_checker.py
```diff
@@ -0,0 +1,295 @@
...
+            print(f"  {i}. {issue.description}")
+            print(f"     Fix: {issue.fix}")
+    print()
+
+
+# ---------------------------------------------------------------------------
+# Issue analysis (pure logic — no I/O)
+# ---------------------------------------------------------------------------
+
+def _analyze_issues(
```
odaiodeh: why this is a standalone function, no class, where it belongs, why its name start with _ ? _ means private/protected in python

local-development/devtool/services/health_checker.py
```diff
@@ -0,0 +1,295 @@
...
+                affected=[sh.name],
+            ))
+
+    return issues
+
+
+# ---------------------------------------------------------------------------
+# Helpers (pure logic)
+# ---------------------------------------------------------------------------
+
```
odaiodeh, line 129: same comment about these, maybe you need util file for this tyoe if services?

local-development/devtool/services/dotenv/common.py
```diff
@@ -0,0 +1,43 @@
+"""Shared constants, types, and helpers for the env package."""
+
+from __future__ import annotations
+
+import re
+from enum import Enum
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from devtool.domain.models import Service
```
odaiodeh: please remove the TYPE_CHECKING, and check that in all local dev there is no TYPE CHECKING

odaiodeh (approved):

yhabushi79 (approved):
well done!

Nirsisr (approved):

PR #148, opened 2026-04-17
Genie 1471/epic/authentication layer infrastructure

Lina-AbuYousef:
## Release Notes

**New Features**
* Added comprehensive OAuth2 authentication system with token storage, automatic refresh, and credential management.
* Introduced authentication endpoints for token exchange, status checking, and client configuration.
* Added pre-configured OAuth integrations for GitHub, Google, and Jira.
* Enabled authentication state detection and dynamic client registration support.

**UI Updates**
* Added authentication status indicators and sign-in interface in workspace.
* Implemented OAuth callback handling and token validation display.


multi-agent/adapters/outbound/redis/auth_pending_store.py
```diff
@@ -0,0 +1,47 @@
+"""
+RedisPendingStore — implements :class:`PendingStore` using Redis.
+
```
sfiresht, line 3: Note that in the sso-redis PR I've created a new adapter/port in the global_utils for all modules to use.
this might save you code and make it more efficient.
let's talk about it once my PR is pushed.

shared-resources/identity/config/app_config.py
```diff
@@ -17,3 +17,8 @@ class AppConfig(SharedConfig):
     frontend_url: str = "http://localhost:5000"    # session_cookie_secure=True
     backend_env: str = "development"
 
+
+    # Multi-agent connection
```
sfiresht, line 23: 1. this shouldn't be here as we can get this from an existing configmap
2. why is it needed in the first place?

shared-resources/sso-backend/endpoints/__init__.py
```diff
@@ -1,12 +1,13 @@
 from endpoints.protected_routes import protected_bp
```
sfiresht: note that the sso-backend was refactored and we'll need to adjust this code to fit it and be included in the new "service"

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
```diff
@@ -0,0 +1,123 @@
...
+from flask import Blueprint, request, make_response
+
+from config.app_config import AppConfig
+
+logger = logging.getLogger(__name__)
+
+credentials_bp = Blueprint("credentials", __name__)
+
+_POPUP_CLOSE_TEMPLATE = """\
+<!DOCTYPE html>
```
MayaCrmi, line 28: can you attach an image for this? i've never seen this kind of popup approach. what makes us need to use this type of display?

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
```diff
@@ -0,0 +1,123 @@
...
+        })
+
+    if not code or not state:
+        return _popup_response({
+            "type": "credentials_callback",
+            "success": False,
+            "error": "Missing code or state",
+        })
+
+    multi_agent_url = f"http://{config.multiagent_host}:{config.multiagent_port}/api/credentials/exchange"
```
MayaCrmi, line 96: can this be added to one of our config files instead of created here?

multi-agent/bootstrap/container.py
```diff
@@ -108,9 +127,94 @@ def __init__(self, cfg: AppConfig):
...
+        http_client = HttpxAuthClient()
+        self.token_store = MongoTokenStore(
+            mongodb_ip=cfg.mongodb_ip,
+            mongodb_port=cfg.mongodb_port,
+            db_name=cfg.mongo_db,
+            coll_name=cfg.auth_tokens_coll,
+        )
+
+        redis_url = get_redis_url()
+        pending_store = None
```
MayaCrmi, line 145: I see this stays None unless the redis_url isn't None, but we still use it in different parts later - will all of the calls to it work even if it's None? for example on line 175 self.auth_exchange_service = OAuth2ExchangeService(
            state_manager=state_manager,
            pending_store=pending_store,
            token_store=self.token_store,
            scheme=oauth2_scheme,
        )

multi-agent/adapters/outbound/mongo/auth_token_repository.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class MongoTokenStore(TokenStore):
+
+    def __init__(
+        self,
+        mongodb_ip: str = "127.0.0.1",
+        mongodb_port: int = 27017,
+        db_name: str = "unifai",
+        coll_name: str = "auth_tokens",
```
MayaCrmi: are we saving this in a hash or somethign like that? assuming it's actual tokens we're saving. can you expand in general a bit about its lifetime? re they removed/replaced/something like that?

multi-agent/adapters/outbound/redis/auth_pending_store.py
```diff
@@ -0,0 +1,47 @@
...
+    def save(self, pending: PendingAuth) -> None:
+        key = f"{_PREFIX}{pending.state_hash}"
+        ttl = max(
+            int((pending.expires_at - datetime.now(timezone.utc)).total_seconds()),
+            60,
+        )
+        payload = pending.model_dump(mode="json")
+        self._redis.setex(key, ttl, json.dumps(payload))
+
+    def consume(self, state_hash: str) -> Optional[PendingAuth]:
```
MayaCrmi: a short explanation for the meaning of each part for the structure and the flow would be really helpful! :)

multi-agent/lib/mas/actions/providers/mcp/validate_connection/validate_connection.py
```diff
@@ -1,158 +1,204 @@
...
-        default=McpTransportType.STREAMABLE_HTTP,
-        description="Transport protocol for MCP server communication"
-    )
-    additional_headers: Dict[str, Any] = Field(
-        default_factory=dict,
-        description="Additional HTTP headers to include in MCP server requests"
-    )
+    user_id: str = Field(default="")
+    server_identifier: str = Field(default="")
+    transport_type: McpTransportType = Field(default=McpTransportType.STREAMABLE_HTTP)
```
MayaCrmi, line 34: is the description for these fields set in a different part now? why do we remove it?

multi-agent/lib/mas/actions/providers/mcp/validate_connection/validate_connection.py
```diff
@@ -1,158 +1,204 @@
...
-        AFTER the async code has finished executing.
-        
-        Because this error occurs outside the async execution context, it cannot
-        be caught inside the async execute() method. We must catch it here at
-        the sync boundary.
-        
-        This is a workaround for a third-party library bug, not a flaw in our
-        AsyncBridge implementation.
-        """
+    def execute_sync(self, input_data, context=None):
```
MayaCrmi, line 71: same, is the comment here not relevant anymore?

multi-agent/lib/mas/core/auth/schemes/api_key/adapter.py
```diff
@@ -0,0 +1,67 @@
...
+        headers.update({
+            "Content-Type": "application/json",
+            "Accept": "application/json, text/event-stream",
+        })
+        try:
+            async with httpx.AsyncClient(timeout=8.0) as client:
+                resp = await client.post(
+                    server_url,
+                    json={
+                        "jsonrpc": "2.0", "id": 1,
```
MayaCrmi: is this json staying the same all the ti,e? how do we make sure to change it easily?

multi-agent/lib/mas/core/auth/service.py
```diff
@@ -0,0 +1,299 @@
...
+            server_identifier=server_identifier,
+            config=config,
+        )
+
+    def bind_lazy(
+        self,
+        ctx_holder: Any,
+        server_identifier: str,
+    ) -> Optional[AuthCredential]:
+        """Create a credential handle with deferred user_id resolution.
```
MayaCrmi: when would bind_lazy be called instead of the bind?

multi-agent/lib/mas/core/enums.py
```diff
@@ -1,31 +1,32 @@
...
-
-class SystemNodeType(str, Enum):
-    """Node types that stay inline in blueprints (never saved as resources)."""
-    USER_QUESTION = "user_question_node"
-    FINAL_ANSWER = "final_answer_node"
-    
-    @classmethod
-    def values(cls) -> frozenset:
-        """All system node type values as strings."""
+from enum import Enum
```
MayaCrmi, line 1: did something change here? why does it look like the entire file was changed? it look the same

multi-agent/lib/mas/elements/auths/jira_oauth/spec/spec.py
```diff
@@ -0,0 +1,17 @@
+from mas.core.enums import ResourceCategory
```
MayaCrmi, line 1: so i a general question if i understand correcrtly - for each mcp we want to support we'll need to create these 4-5 files (spec, init, config, etc...)? if people bring other mcps it will work with a general file?

Nirsisr:
> * Implemented

@Lina-AbuYousef Please for the next time. When you put such a big EPIC under CR put your own words explanation for it. I know we had a long call 1 on 1 tmr going over the details but for everyone else sake.

CodeRabbit summarizes it in very high level which is great but it will be impossible to start CR it without some more information. Regard the GUI (MCP inventory type) please align with the changes we talked about yesterday. 
User will need to chose the type of authentication from dropdown whether it's bearer token OR 2 Factor Auth, and if he chose 2 Factor Auth he might chose optional field Type like Google/Jira when you show the new dropdown item in the list per those items if it doesn't have auto-registry.

If he chose the other type he will see the bearer token field rendered.  Currently when we have no separation it's really messy. 

+ in the past if user first put the token and than the MCP_URL the action VALIDATE API return VALID same behavior need to happen along your changes. Make sure we are not breaking any previous behavior here.

multi-agent/lib/mas/elements/auths/oauth_client/config.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Literal, Optional
+
+from pydantic import Field
+
+from mas.core.field_hints import ActionHint, HintType
+from mas.elements.auths.common.base_config import AuthBaseConfig
+from .identifiers import Identifier
+
+
+class OAuthClientConfig(AuthBaseConfig):
```
MayaCrmi, line 18: i assume this is the said general file i askes about before? like for mcps we don't support in this new way?

multi-agent/lib/mas/elements/auths/oauth_client/oauth_client_factory.py
```diff
@@ -0,0 +1,39 @@
...
+from mas.elements.common.exceptions import PluginConfigurationError
+
+from .config import OAuthClientConfig
+from .identifiers import Identifier
+from .oauth_client_instance import OAuthClientInstance
+
+logger = logging.getLogger(__name__)
+
+_ACCEPTED_TYPES = frozenset({
+    "oauth_client", "google_oauth", "github_oauth", "jira_oauth",
```
MayaCrmi, line 22: haha i assuem this answers my question :D is there a general config file for this instea dof having it in this file?

multi-agent/lib/mas/elements/providers/mcp_server_client/mcp_provider.py
```diff
@@ -171,7 +193,7 @@ def refresh_tools(self) -> None:
         with get_async_bridge() as bridge:
             bridge.run(self._initialize_tools())
 
-    def clone(self) -> "McpProvider":
+    def clone(self) -> McpProvider:
```
MayaCrmi, line 195: why did we need to remove the. "" ? how was this changed? does it affect amything?

multi-agent/lib/mas/elements/providers/mcp_server_client/validator.py
```diff
@@ -17,52 +22,40 @@
...
+
+logger = logging.getLogger(__name__)
+
+_MCP_INIT_BODY = {
+    "jsonrpc": "2.0",
+    "id": 1,
+    "method": "initialize",
+    "params": {
+        "protocolVersion": "2024-11-05",
+        "capabilities": {},
```
MayaCrmi, line 37: is this the same as the json i asked about before? can we pout it someplace else instead of having to manage it in both parts?

multi-agent/lib/mas/elements/tools/mcp_proxy/mcp_proxy_tool.py
```diff
@@ -74,43 +82,66 @@ def run(self, *args: Any, **kwargs: Any) -> Any:
...
-        
-        # 2) Use the fresh client for all operations
+
         async with client:
-            # 3) Fetch schema once if not already
             if not self._schema_initialized:
                 await self._ensure_tool_info(client)
 
-            # 4) Validate + prune arguments
             mcp_args = self._prepare_arguments(kwargs)
```
MayaCrmi, line 118: why did we remove all of the comments here? are they not relevant anymore? just for it to stay clear and sorted

shared-resources/identity/config/app_config.py
Nirsisr, line 23: @Lina-AbuYousef The SSO/identity service should be consumed by the other services.

He must be completely de-coupled from the APP. I am not sure why you put those multi-agent parmeters here, but it shouldn't be here. I guess it might be realted to the face that the SSO need to re-direct some req for the 2 factor Authentication to the multi-agent if that's the case we should find another way.

Maybe PUSH NOTIFICATIONS queue that multi-agent listens too with updated on returned request into the SSO pod from external websites (the ones that we need for the 2 factor auth). Or try to think about a way that doesn't require to config it here.

ui/client/src/components/agentic-ai/workspace/AuthSelector.tsx
```diff
@@ -0,0 +1,248 @@
...
+type AuthState = 'idle' | 'checking' | 'authenticated' | 'requires_consent' | 'error';
+
+interface OptionAuthState {
+  status: AuthState;
+  authUrl: string | null;
+  message: string;
+  serverIdentifier: string;
+}
+
+const PROVIDER_ICONS: Record<string, React.ReactNode> = {
```
MayaCrmi: maybe this should be in a different file if we'll want to use them in the future for reference in templates for excample or something to note which provider it is

ui/client/src/components/agentic-ai/workspace/AuthSelector.tsx
```diff
@@ -0,0 +1,248 @@
...
+            status: 'error',
+            message: event.data.error || 'Authentication failed',
+          });
+        }
+        activeOptionRef.current = null;
+      }
+    };
+    window.addEventListener('message', onMessage);
+    return () => window.removeEventListener('message', onMessage);
+  }, [refOptions, checkAuthStatus]);
```
MayaCrmi, line 149: if we update checkAuthStatus here and it's in the dep list wouldn't this be called too much?

ui/client/src/components/agentic-ai/workspace/AuthSelector.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+    updateAuthState(option.rid, {
+      status: 'checking',
+      authUrl: null,
+      message: '',
+      serverIdentifier,
+    });
+
+    try {
+      const response = await axios.post('/actions/action.execute', {
```
MayaCrmi: can we change this to be a call like executeAction() with the actual axios call in a http file, like in toher parts?

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
```diff
@@ -0,0 +1,123 @@
+"""
+Credentials Callback Relay
+
+Receives OAuth callbacks from external authorization servers (GitHub, Atlassian, etc.)
+and relays the authorization code + state to the multi-agent backend for token exchange.
+
+The SSO pod is publicly accessible, so it can receive OAuth redirects.
```
Nirsisr, line 7: @Lina-AbuYousef That's a good point. Is the only reason you direct the Oauth redirects to the SSO pod is since is publicly accessible? Or you are using some functionality from within the SSO pod for those re-directions as well?

Maybe in another ticket we need to come up with a proxy server for that purposes. Since it's not making sense using the SSO pod for that re-directions since it's the only service we have which is publicly available.

@sfiresht we need to make the general backend we have like a proxy-server with external route accessible for outside request, so lina can leverage this one and re-directs her request from there to any of our services. That way the general backend (proxy server) also aware for all the services below (it's part of the configuration) and we don't need to hard coded the MULTI-AGENT service as part of the SSO pod.

ui/client/src/components/agentic-ai/workspace/ElementForm.tsx
```diff
@@ -16,6 +16,7 @@ import {
 } from "../../../types/workspace";
```
MayaCrmi, line 16: i think the changes in the file aren't needed (just a newline and removed comments that i don't think need to be removed)

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -27,8 +28,15 @@ interface FieldValidationProps {
   /** All current config field values, used to resolve dependencies for validation actions */
   configValues?: Record<string, any>;
   onValidationChange: (fieldName: string, isValid: boolean, itemResults?: ItemValidationResult[]) => void;
+  onInputChange?: (field: string, value: any) => void;
 }
 
+// Auth-related response statuses
+const AUTH_STATUSES = new Set([
```
MayaCrmi, line 37: is there a different filke this can be added to? with other statuses etc?

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
```diff
@@ -0,0 +1,123 @@
...
+    multi_agent_url = f"http://{config.multiagent_host}:{config.multiagent_port}/api/credentials/exchange"
+    try:
+        resp = requests.post(
+            multi_agent_url,
+            json={"code": code, "state": state},
+            headers={
+                "Content-Type": "application/json",
+            },
+            timeout=15,
+        )
```
Nirsisr, line 105: This part as mentioned above. We must deal from our higher level backend, proxy-server for all of the requests in UnifAI.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -82,31 +94,18 @@ export const FieldValidation: React.FC<FieldValidationProps> = ({
     });
   }, [fieldValue, validationHint?.dependencies, fieldName, configValues]);
```
MayaCrmi, line 104: same comment about comments and explanations removed

odaiodeh:
@Lina-AbuYousef 
# Auth Layer Review: Current Design vs. Proposed Design

> Design review for **GENIE-1471 / Authentication Layer Infrastructure**.
> This document captures the smells in the current implementation, the proposed cleaner design, and a side-by-side comparison.

---

## Table of Contents

1. [Smells in the Current Design](#smells-in-the-current-design)
2. [Side-by-Side: Current vs. Proposed](#side-by-side-current-vs-proposed)
   - [Components](#components)
   - [Architecture](#architecture)
   - [The Port (Strategy Interface)](#the-port)
   - [The Action Layer](#the-action-layer)
   - [AuthChallenge — Scheme-Agnostic Responses](#authchallenge)
   - [The Unified Flow](#the-unified-flow)
   - [Consumer Side](#consumer-side)
   - [Storage Ports](#storage-ports)
3. [The Litmus Test: Adding SAML](#the-litmus-test-adding-saml)
4. [Summary Table](#summary-table)

---

## Smells in the Current Design

### Smell 1: OAuth2 Leaks Upward

```
  ┌──────────────────────────────────────────────────────────┐
  │  AuthenticateAction                                      │
  │                                                          │
  │  depends on:                                             │
  │    AuthService            ← generic ✓                    │
  │    OAuth2LoginService     ← OAuth2-specific ✗            │
  │                                                          │
  │  execute():                                              │
  │    token = auth_service.get_valid_token()                 │
  │    if not token:                                         │
  │      url = oauth2_login_service.build_login_url()  ← ✗   │
  │      return {status: "requires_consent", url}            │
  └──────────────────────────────────────────────────────────┘

  If you add SAML or API Key onboarding, you must EDIT this action.
  That violates Open/Closed Principle.
```

### Smell 2: Three Peer Services, No Single Owner

```
  CURRENT:                          WHO OWNS THE AUTH LIFECYCLE?

  AuthService ─────────────── "I get tokens and build headers"
  OAuth2LoginService ─────── "I start OAuth logins"
  OAuth2ExchangeService ──── "I finish OAuth callbacks"

  Answer: nobody owns it all. The lifecycle is split across three
  services at the same level. The container and the actions must
  know about all three and wire them together manually.
```

### Smell 3: One Interface Does Two Jobs

```
  ┌────────────────────────────────────────┐
  │  OAuth2Scheme                          │
  │                                        │
  │  Runtime (using tokens):               │
  │    build_headers()                     │  ← about USING credentials
  │    validate()                          │
  │    attempt_recovery()                  │
  │                                        │
  │  Onboarding (getting tokens):          │
  │    build_login_context()               │  ← about ACQUIRING credentials
  │    exchange_credentials()              │
  │    register_client()                   │
  └────────────────────────────────────────┘

  Not a violation by itself, but combined with Smell 2 it means
  the login/exchange logic exists BOTH inside OAuth2Scheme AND
  inside OAuth2LoginService/OAuth2ExchangeService — unclear who
  is responsible for what.
```

### Smell 4: Schemes Know About MCP

```
  OAuth2Scheme.validate():
    POST server_url with JSON-RPC {"method": "initialize"} ← MCP protocol

  ApiKeyScheme.validate():
    POST server_url with JSON-RPC {"method": "initialize"} ← MCP protocol

  Auth strategies should not know what protocol the resource server
  speaks. MCP is a consumer detail, not an auth detail.
```

### Smell 5: Wrapper Around a Wrapper

```
  MCP Provider → OAuthClientInstance → BoundCredential → AuthService

  OAuthClientInstance just delegates every call to BoundCredential.
  It adds no logic. It exists because the element factory creates
  hollow instances that get wired later. Two layers of indirection
  for zero benefit.
```

### Smell 6: Split Construction

```
  Step 1 (Element layer):
    OAuthClientFactory.create()
      → OAuthClientInstance()        ← created with NO auth_service

  Step 2 (Session builder):
    ProviderBuilder._extra_kwargs()
      → auth_service.bind_lazy()     ← auth injected HERE, separately

  The object is constructed in two places. If either step is missed
  or ordered wrong, you get a hollow instance that silently returns
  empty headers.
```

---

## Side-by-Side: Current vs. Proposed

### Components

```
  CURRENT (15 concepts)                 PROPOSED (10 concepts)
  ─────────────────────                 ──────────────────────

  AuthService                           AuthService
  OAuth2LoginService                    (gone → inside OAuth2Strategy)
  OAuth2ExchangeService                 (gone → inside OAuth2Strategy)
  OAuthStateManager                     (gone → internal to OAuth2Strategy)
  SchemeRegistry                        AuthStrategyRegistry
  AuthScheme (port)                     AuthStrategy (port)
  InteractiveAuthScheme (port)          (gone → merged into AuthStrategy)
  OAuth2Scheme                          OAuth2Strategy
  ApiKeyScheme                          ApiKeyStrategy
  BoundCredential                       AuthHandle
  AuthCredential (lazy)                 LazyAuthHandle
  OAuthClientInstance                   (gone)
  AuthDetector                          (gone → AuthService.discover())
  DetectionStrategy (port)              AuthDiscovery (port)
  OAuth2DetectionStrategy               OAuth2Discovery
```

### Architecture

```
  ═══════════════════════════════════════════════════════════════
  CURRENT
  ═══════════════════════════════════════════════════════════════

  Action ──► AuthService ─────────────────► SchemeRegistry
    │                                         ├─ OAuth2Scheme
    │                                         └─ ApiKeyScheme
    │
    ├──► OAuth2LoginService ──► OAuth2Scheme
    │         │
    │         ├──► OAuthStateManager
    │         ├──► PendingStore
    │         └──► ClientConfigStore
    │
    └─────────────────── (separate service, separate wiring)

  Endpoint ──► OAuth2ExchangeService ──► OAuth2Scheme
                    │
                    ├──► OAuthStateManager
                    ├──► PendingStore
                    └──► TokenStore

  3 services.  Actions import 2.  Endpoints import 1.
  Adding SAML = new service + change action + change container.


  ═══════════════════════════════════════════════════════════════
  PROPOSED
  ═══════════════════════════════════════════════════════════════

  Action ──► AuthService ──► AuthStrategyRegistry
                │                ├─ OAuth2Strategy (self-contained)
  Endpoint ─────┘                ├─ ApiKeyStrategy (self-contained)
                                 └─ SamlStrategy   (future, same shape)

  1 service.  Everyone imports 1 thing.
  Adding SAML = new strategy + register. Done.
```

### The Port

```
  ═══════════════════════════════════════════════════════════════
  CURRENT: Two ports, partial overlap
  ═══════════════════════════════════════════════════════════════

  AuthScheme                          InteractiveAuthScheme
  ┌──────────────────────┐            ┌──────────────────────────┐
  │ scheme_type           │            │ (extends AuthScheme)      │
  │ build_headers()       │            │ build_login_context()     │
  │ validate()            │ ← MCP!     │ exchange_credentials()    │
  │ attempt_recovery()    │            │ register_client()         │
  └──────────────────────┘            └──────────────────────────┘

  ApiKeyScheme implements AuthScheme only.
  OAuth2Scheme implements InteractiveAuthScheme.
  The action must check which type it has and branch.


  ═══════════════════════════════════════════════════════════════
  PROPOSED: One port, complete contract
  ═══════════════════════════════════════════════════════════════

  AuthStrategy
  ┌──────────────────────────────────────┐
  │  scheme_type: str                    │
  │                                      │
  │  build_headers(credential) → headers │  ← runtime
  │  refresh(credential, config) → token │  ← runtime
  │                                      │
  │  initiate(config) → AuthChallenge     │  ← onboarding
  │  complete(data, context) → TokenSet  │  ← onboarding
  └──────────────────────────────────────┘

  Every strategy implements all four methods.
  No type-checking. No branching. Pure polymorphism.
```

### The Action Layer

```
  ═══════════════════════════════════════════════════════════════
  CURRENT
  ═══════════════════════════════════════════════════════════════

  class AuthenticateAction:
      def __init__(self,
          auth_service,              ← generic
          oauth2_login_service,      ← OAuth2-specific
      ): ...

      async def execute(self, input):
          token = await self.auth_service.get_valid_token(...)
          if token:
              return {"status": "authenticated"}

          config = await self.auth_service.get_client_config(...)
          url = await self.oauth2_login_service.build_login_url(...)
                        ^^^^^^^^^^^^^^^^^^^^^^^^
                        knows it's OAuth2

          return {"status": "requires_consent", "url": url}


  ═══════════════════════════════════════════════════════════════
  PROPOSED
  ═══════════════════════════════════════════════════════════════

  class AuthenticateAction:
      def __init__(self,
          auth_service,              ← only dependency
      ): ...

      async def execute(self, input):
          status = await self.auth_service.check_status(...)
          if status == AUTHENTICATED:
              return {"status": "authenticated"}

          challenge = await self.auth_service.initiate(...)
                            ^^^^^^^^^^^^^^^^^^^^
                            scheme-agnostic

          return challenge.to_response()
          # Could be {"status": "requires_consent", "url": "..."}
          # Could be {"status": "requires_input", "fields": [...]}
          # Could be anything. Action doesn't care.
```

### AuthChallenge

What strategies return to the UI — scheme-agnostic, polymorphic.

**Why "Challenge"?** In auth protocols, a *challenge* is what the server sends
back when the client hasn't proven its identity yet — "prove who you are."
RFC 7235 (HTTP Authentication) formally defines this as the `WWW-Authenticate`
challenge. The name fits every variant: "we challenge you to consent via OAuth,"
"we challenge you to provide an API key," "we challenge you to enter a device
code." Alternatives like "Prompt" imply asking for input (doesn't fit the
redirect case), and "Requirement" is too passive. Challenge is precise,
protocol-standard, and reads naturally in code:
`challenge = await auth_service.initiate(...)` — "we initiated auth, got a
challenge back."

```
  ═══════════════════════════════════════════════════════════════
  CURRENT: OAuth2-specific response shape
  ═══════════════════════════════════════════════════════════════

  {"status": "requires_consent", "authorization_url": "https://..."}

  Hardcoded in the action. Only works for OAuth2.


  ═══════════════════════════════════════════════════════════════
  PROPOSED: Scheme-agnostic AuthChallenge
  ═══════════════════════════════════════════════════════════════

  AuthChallenge
    │
    ├── AuthChallenge.consent(url, flow_id)
    │   {"type": "consent", "url": "https://...", "flow_id": "abc"}
    │   Strategy: OAuth2, SAML
    │   UI: opens popup window
    │
    ├── AuthChallenge.collect(fields, flow_id)
    │   {"type": "collect", "fields": [{"name":"api_key","secret":true}], "flow_id": "abc"}
    │   Strategy: ApiKey, BasicAuth
    │   UI: renders input form
    │
    └── AuthChallenge.device(uri, code, flow_id)
        {"type": "device", "uri": "https://...", "code": "ABCD-1234", "flow_id": "abc"}
        Strategy: DeviceFlow (future)
        UI: shows code + link
```

### The Unified Flow

```
  ═══════════════════════════════════════════════════════════════
  CURRENT: Different paths for different sources
  ═══════════════════════════════════════════════════════════════

  Auth Element (Google)      MCP Discovery
       │                          │
       ▼                          ▼
  AuthenticateAction         ValidateConnectionAction
       │                          │
       ▼                          ▼
  OAuth2LoginService         AuthDetector
       │                          │
  (different path)           (different path)


  ═══════════════════════════════════════════════════════════════
  PROPOSED: One path regardless of source
  ═══════════════════════════════════════════════════════════════

  Auth Element (Google)      MCP Discovery           API Key
       │                          │                      │
       │ scheme_type="oauth2"     │ detect → "oauth2"    │ scheme_type="api_key"
       │ server="google.com"      │ server="mcp.io"      │ server="api.io"
       │ client_id=known          │ client_id=from DCR   │
       │                          │                      │
       └──────────┬───────────────┘──────────────────────┘
                  │
                  ▼
        AuthService.initiate(user, server, scheme_type, config)
                  │
                  ▼
        AuthStrategyRegistry.get(scheme_type)
                  │
          ┌───────┼────────┐
          ▼       ▼        ▼
      OAuth2       ApiKey      (any future)
      Strategy     Strategy    Strategy
          │            │           │
          ▼            ▼           ▼
      AuthChallenge AuthChallenge AuthChallenge
      .consent()    .collect()    .(whatever)
```

### Consumer Side

```
  ═══════════════════════════════════════════════════════════════
  CURRENT: Three layers of wrapping
  ═══════════════════════════════════════════════════════════════

  McpProvider
    → OAuthClientInstance                    ← wrapper #1
       → BoundCredential                    ← wrapper #2
          → AuthService.build_headers()


  ═══════════════════════════════════════════════════════════════
  PROPOSED: Direct handle
  ═══════════════════════════════════════════════════════════════

  McpProvider
    → AuthHandle.get_headers()              ← one object
       → AuthService.build_headers()
```

### Storage Ports

```
  ═══════════════════════════════════════════════════════════════
  CURRENT                              PROPOSED
  ═══════════════════════════════════════════════════════════════

  TokenStore                           CredentialStore
    what it stores: tokens               what it stores: credentials
    name suggests narrow scope           name matches actual scope

  ClientConfigStore                    ServerConfigStore
    what it stores: client configs       what it stores: auth server configs
    name is OAuth2-centric              name is protocol-agnostic

  PendingStore                         FlowStateStore
    what's pending? unclear              stores in-progress flow state
    name is vague                        name is self-describing

  PendingAuth                          FlowState
    "auth that is pending"?              state of an auth flow in progress
```

---

## The Litmus Test: Adding SAML

```
  ═══════════════════════════════════════════════════════════════
  CURRENT: What you must touch
  ═══════════════════════════════════════════════════════════════

  1. Create SamlScheme (implements InteractiveAuthScheme)
  2. Create SamlLoginService (new peer service)
  3. Create SamlExchangeService (new peer service)
  4. Edit AuthenticateAction to import SamlLoginService
  5. Edit container.py to wire 2 new services
  6. Edit container.py to inject into action
  7. Edit action execute() to branch on SAML vs OAuth2

  7 changes across 4 files. 3 new files.


  ═══════════════════════════════════════════════════════════════
  PROPOSED: What you must touch
  ═══════════════════════════════════════════════════════════════

  1. Create SamlStrategy (implements AuthStrategy)
  2. Register it: registry.register(SamlStrategy())

  2 changes. 1 new file. Nothing else touched. Ever.
```

---

## Summary Table

| | Current | Proposed |
|---|---|---|
| **Top-level services** | 3 (`AuthService`, `OAuth2LoginService`, `OAuth2ExchangeService`) | 1 (`AuthService`) |
| **Ports for strategies** | 2 (`AuthScheme`, `InteractiveAuthScheme`) | 1 (`AuthStrategy`) |
| **Registries** | 1 (schemes only, flows are separate services) | 1 (`AuthStrategyRegistry`) |
| **Wrapper layers** | 3 (`OAuthClientInstance` → `BoundCredential` → `AuthService`) | 1 (`AuthHandle` → `AuthService`) |
| **Action dependencies** | 2 (`AuthService` + `OAuth2LoginService`) | 1 (`AuthService`) |
| **Total concepts** | 15 | 10 |
| **To add a new auth method** | 7 changes, 4 files | 2 changes, 1 file |
| **Protocol coupling** | Schemes validate via MCP JSON-RPC | Schemes know nothing about consumers |
| **Naming** | Mixed (OAuth2-centric names in generic ports) | Consistent (scheme-agnostic everywhere) |

ui/client/src/components/agentic-ai/workspace/AuthSelector.tsx
```diff
@@ -0,0 +1,248 @@
...
+        uid: 'auth.authenticate',
```
Nirsisr: @Lina-AbuYousef See what we did in FieldValidation or any other component we trigger actions from there. 

We are not hard coding the UID of the ACTIONS we are taking it from the ACTION it-self, it should be generic for future usage.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -272,16 +249,70 @@ export const FieldValidation: React.FC<FieldValidationProps> = ({
...
-        isValid: false,
-        message: errorMessage
-      });
-
+      setValidationState({ isValidating: false, isValid: false, message: errorMessage });
       onValidationChange(fieldName, false);
     }
   };
 
+  const handleAuthResponse = useCallback((data: any) => {
```
MayaCrmi, line 301: could this and maybe other logics be extracted to a hook or something? since this file is for all validation and this is specific for the new auths category

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -308,43 +338,88 @@ export const FieldValidation: React.FC<FieldValidationProps> = ({
...
   if (!useApiHint && !validationAction) {
     return null;
   }
   if (useApiHint && !validationHint.endpoint) {
     return null;
   }
 
+  // ── Auth-aware rendering ──
+
+  if (authStatus === 'authenticated') {
```
MayaCrmi: same comment about general code here

MayaCrmi (commented):
alufa! great job here

ui/client/src/components/agentic-ai/workspace/CategorySidebar.tsx
```diff
@@ -41,6 +43,7 @@ const getCategoryIcon = (category: string) => {
 const getCategoryDisplayName = (category: string) => {
   const nameMap: { [key: string]: string } = {
     nodes: 'Agents',
+    auths: 'Auths',
```
Nirsisr, line 46: @Lina-AbuYousef The plan is to have Auths elements in MAS but to hide them from GUI? 

Or we even going not to have them at all, and they just configured as fields into the MCP.

ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
```diff
@@ -307,7 +311,8 @@ export const FieldPopulation: React.FC<FieldPopulationProps> = ({
 
     const response = await axios.post('/actions/action.execute', {
       uid: populateAction.uid,
-      inputData
+      inputData,
+      userId,
```
Nirsisr: Lina's what changed in the populate action that currently it require userId?

If you can explain here the new populate action you created which require it, will appreciate that.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -49,8 +61,14 @@ export const FieldValidation: React.FC<FieldValidationProps> = ({
...
   });
 
+  // Auth-specific state
+  const [authUrl, setAuthUrl] = useState<string | null>(null);
+  const [authStatus, setAuthStatus] = useState<string | null>(null);
+  const [authMessage, setAuthMessage] = useState<string | null>(null);
+
   const validationTimeoutRef = useRef<NodeJS.Timeout | null>(null);
   const lastValidatedKeyRef = useRef<string | null>(null);
+  const popupRef = useRef<Window | null>(null);
```
Nirsisr, line 74: Lina we had here very important comments above validationKey.

```
 /**
   * Creates a stable validation key that combines the field value and all dependency values.
   * This key is used to determine if validation should be re-triggered.
   * When either the field value or any dependency changes, this key will change.
   */

```

Please make sure to bring all of those back.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Nirsisr, line 1: @Lina-AbuYousef You remove a lot of important comments probably since Cursor did it or something in this file. Please bring all of them back. Each of them where generated separately after thinking carefully on each of the steps here. We can't have them gone.

**Please double-check Cursor not removing us important stuff from our code.**

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -272,16 +249,70 @@ export const FieldValidation: React.FC<FieldValidationProps> = ({
...
+      setValidationState({ isValidating: false, isValid: false, message: data.message || '' });
+      onValidationChange(fieldName, false);
+    } else {
+      setAuthUrl(null);
+      setAuthStatus(status);
+      setAuthMessage(message);
+      setValidationState({ isValidating: false, isValid: false, message });
+      onValidationChange(fieldName, false);
+    }
+  }, [fieldName, onValidationChange]);
```
Nirsisr: lina's please write it more elegantly with Switch/Case for each of the cases which trigger an helper functions which setX setY setZ etc. with the relevant keys.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Nirsisr: @Lina-AbuYousef I am not sure how complex is it, but if it's possible. 

It will be much more clear for us taking the entire 2 Factor Auth logic, including the rendering part with the entire relevant state:
`// ── Auth-aware rendering ──`

Into sub-component as FieldValidationTwoFactorAuth, so the entire logic will live there if we need to adjust in the future. You have here a lot of extra logic like evenListeners, re-directions, different Errors and a lot of more to handle. It's better we export this logic outside of FieldValidation to avoid confusion.

Nirsisr (changes requested):

Nirsisr:
@Lina-AbuYousef I know we discuss it a bit in the call but writing it here:

`client_configs is keyed by server_identifier only (not per user/tenant);  which is good for shared app credentials, but very risky for multi-tenant isolation.`

For that reason and more, I prefer go with the AuthComponent configured from the GUI by the user, each will need to bring his client etc. (adjusting to our low-code/no-code unifAI approach) instead of hard coding some servers under server_identifier. 

We can have per each user some of pre-defined AuthCreds for google/jira etc. that he can't delete, instead of the user need to create his own we offer some build in by unifAI so he don't need to create his own Client. 

If user put MCP jira/google from the dropdown we can automatically select for him the relevant Auth element, while he will be able to change it to another Auth component available in the list (assuming he created few of them). With that approach is much more aligned to what we have in UnifAI with refs elements and more.

multi-agent/adapters/inbound/flask/endpoints/actions.py
```diff
@@ -79,36 +79,16 @@ def list_actions(category=None, type=None, action_type=None, tags=None):
 @from_body({
     "uid": fields.Str(required=True),
     "input_data": fields.Dict(data_key="inputData", required=False, load_default={}),
-    "context": fields.Dict(required=False, load_default={})
+    "context": fields.Dict(required=False, load_default={}),
```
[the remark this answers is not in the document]
Lina-AbuYousef: Marshmallow's load_default with {} is safe. Unlike Python function default arguments, marshmallow doesn't reuse the default object — it creates a new instance each time it deserializes. The load_default value is used as a template, and marshmallow's Field.load method handles copying internally.

multi-agent/lib/mas/core/enums.py
Lina-AbuYousef: IDK why sometime that's happen when cursor change stuff but the only change is   AUTH = "auths"

multi-agent/lib/mas/elements/auths/jira_oauth/spec/spec.py
Lina-AbuYousef: I think I have to do a session for the team since there is a lot of cases here
but if the mcp supports the auto discovery then no need at all for the auth elements

shared-resources/identity/config/app_config.py
Lina-AbuYousef: Hmm If I got you right, you mean we should have the config saved maybe in the env vars, and to call it smth more general to where we navigate the requests?

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
Lina-AbuYousef: Right now yes this is the case, the sso pod is the only one that publicly accessible, but in the future my vision to have a dedicated microservice that handles all the auth layer and maybe to have for it a router that is not behind a vpn

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
Lina-AbuYousef: I thought about it, hmm like idk if this a common approach to put the whole url in app cofig but maybe you're right and this could be better will check and update

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
Lina-AbuYousef: right, I agree with we , we just need to discus if to do it now or phase 2

ui/client/src/components/agentic-ai/workspace/AuthSelector.tsx
Lina-AbuYousef: agree

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Lina-AbuYousef: done

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Lina-AbuYousef: done, returned comments and created a dedicated component

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Lina-AbuYousef: done

ui/client/src/components/agentic-ai/workspace/FieldPopulation.tsx
Lina-AbuYousef: you need the user id so the multi agent could get the user credentials from the credentials store, we are retrieving them by server id + user id and to check the cred ,  as the multi agent do not have any kind of user session so the ui needs to provide it

ui/client/src/components/agentic-ai/workspace/CategorySidebar.tsx
Lina-AbuYousef: I think in this phase I will hide it since all the required mcp servers by omri was hadnled by auto discovery or the bearer token, but I will keep the code for phase two

ui/client/src/components/agentic-ai/workspace/AuthSelector.tsx
Lina-AbuYousef: done

multi-agent/adapters/outbound/redis/auth_pending_store.py
Lina-AbuYousef: this is the port that redis implement, this port is used in oauth2 for tmp stuff like exchanging the token of the user with the server. 
multi agent in the first call have only a "code" string, it save it in redis and exchange it with a real token

shared-resources/identity/config/app_config.py
Nirsisr: Right like it has to be part of the ENV VARS, regardless I mention that once we have the backend as a proxy server we need to move your piece to there instead of having it it in SSO pod.

shared-resources/identity/adapters/inbound/flask/endpoints/credentials_callback.py
Nirsisr: Completely fair, you can close it.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Nirsisr: 10x lina.

ui/client/src/components/agentic-ai/workspace/AuthFieldRenderer.tsx
```diff
@@ -0,0 +1,224 @@
...
+        setStatus('error');
+        setAuthUrl(null);
+        setMessage(data.message || 'Authentication error');
+        onValidationChange(fieldName, false);
+      }
+    } catch {
+      setStatus('error');
+      setMessage('Failed to check authentication status');
+      onValidationChange(fieldName, false);
+    }
```
Nirsisr, line 120: @Lina-AbuYousef You said on prev comment that you move it to switch/case or something more clean : D

Anyway up to you not super critical but can be nicer & cleaner

ui/client/src/components/agentic-ai/workspace/AuthFieldRenderer.tsx
Nirsisr, line 1: @Lina-AbuYousef The code right now looks X10 better, when you move the entire agent to separated component. Great work !

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
Nirsisr: Lina +1 on that. I see you made changes to FieldValidate and you bring back the comments and all the other comments I had on the CR which is great!

If you can extract as Maya suggested those pieces of FieldValidationOauth to another sub-component, it will be much more cleaner for us.

ui/client/src/components/agentic-ai/workspace/FieldValidation.tsx
```diff
@@ -316,35 +386,45 @@ export const FieldValidation: React.FC<FieldValidationProps> = ({
...
+  if (authStatus) {
+    return (
+      <FieldValidationTwoFactorAuth
+        authStatus={authStatus}
+        authUrl={authUrl}
+        authMessage={authMessage}
+        onRevalidate={handleAuthRevalidate}
+        onAuthError={handleAuthError}
+      />
+    );
```
Nirsisr, line 411: That's exactly was the request which you did. If we can have there also the sub-functions from above like `handleAuthResponse` it's the best.

If it's not right place to locate those, so maybe another sub-components to store those handlers who are responsible for TwoFactorAuth

ui/client/src/contexts/AgenticAIContext.tsx
```diff
@@ -656,7 +656,7 @@ return String(ref);
     setBlueprintValidationStatus(blueprintId, 'loading');
     
     try {
-      const result = await validateBlueprintApi(request);
+      const result = await validateBlueprintApi({ ...request, userId: request.userId || USER_ID });
```
Nirsisr, line 659: @Lina-AbuYousef I understand that if we have MCP with 2Auth you need userId but now we are sending each time USER_ID for the validation endpoint IDK ammm

Like btw we have the USER_ID in the metadata of the blueprint 'created_by' or something like that so we not MUST to send it.

Anyway ignore it for now, once we have Redis with User Store we don't need to send it to the entire APIs, so just put it as a reminder to remove it from here once we have the Redis wrapping the MAS APIs.

Nirsisr (changes requested):
@Lina-AbuYousef Reviewed all for the second-cycle (not including MAS changes).

The improvement you did in GUI make the entire EPIC much much better and solid. 


Please remember our last call notes:
- Get rid of the HIDDEN field with server identity (don't send it via the GUI, find a way to handle it only on the backend side since you already have this stored in Mongo)
- The issue with the face we are storing all the tokens after 'Sign In' even when the MCP element never saved. Or that we keep storing it after removal. (super critical security concern that we must to tackle ASAP). 

Well done linas!

odaiodeh:
@Lina-AbuYousef 

## Auth Layer Code Review — GENIE-1471

### Critical (must fix)

**1. External libraries inside core hexagon**

`core/auth/schemes/oauth2/adapter.py` imports `httpx` and `authlib` — these are external I/O libraries that violate the hexagonal boundary. The container correctly imports from `outbound.auth.oauth2_strategy`, and the `__init__.py` even says "The strategy adapter has moved to `adapters/outbound/auth/`" — but the core copy still contains the full 394-line implementation with all external imports.

Same applies to `core/auth/schemes/api_key/adapter.py` (though it has no external imports, the duplication is confusing).

> **Fix**: Remove the strategy implementations from core, or replace them with re-exports from the outbound adapters. Core should only contain ports, models, configs, and internal utilities — never `import httpx`.

**2. Broken methods in `McpProxyTool`**

- `refresh_schema()` calls `await self._ensure_tool_info()` without the required `client` argument — will crash at runtime.
- `health_check()` references `self.mcp_client` which is never defined in `__init__` — will crash at runtime.

---

### Dependency Inversion issues

**4. `OAuth2Strategy._try_auto_register` reaches into detector internals**

```python
http_client = getattr(self._detector, '_http_client', None)
```

This accesses a private attribute on `AuthDetector`, coupling the strategy to the detector's internal implementation. If the detector restructures, this silently breaks.

It also calls `OAuth2DetectionStrategy._fetch_as_metadata(...)` directly as a static method, bypassing the `DetectionStrategy` abstraction entirely.

> **Fix**: Inject `HttpClient` directly into `OAuth2Strategy` as a constructor parameter. The strategy already needs HTTP access for token refresh and DCR — giving it the HTTP port is cleaner than reaching into another object's internals.

**5. Monkey-patching after construction**

```python
self.resources_service._auth_service = self.auth_service
self.blueprint_service._auth_service = self.auth_service
```

Setting private attributes post-construction breaks encapsulation and deviates from the constructor injection pattern used everywhere else in the container.

> **Fix**: Accept `auth_service` as an optional constructor parameter in `ResourcesService` and `BlueprintService`.

---

### Code quality

**6. Missing enums for discriminator strings**

`TokenStatus` is properly an enum, but several other discriminators used throughout the layer are plain strings:

| Value | Used as | Should be |
|---|---|---|
| `"consent"` / `"collect"` / `"device"` | `challenge_type` | `ChallengeType` enum |
| `"oauth2"` / `"api_key"` | `scheme_type` | `SchemeType` enum |
| `"oauth2"` / `"saml"` | `protocol_type` | `ProtocolType` enum |

Enums catch typos at import time rather than as silent runtime bugs.

**7. Loose typing in key places**

| Location | Current | Should be |
|---|---|---|
| `AuthHandle.__init__(user_id: Any)` | `Any` | `Union[str, ExecutionContextHolder]` |
| `OAuth2Strategy.__init__(client_config_store: Optional[Any])` | `Any` | `Optional[ServerConfigStore]` |
| `OAuth2Strategy.__init__(detector: Optional[Any])` | `Any` | `Optional[AuthDetector]` |
| `GetToolsNamesAction.__init__(auth_service=None)` | no annotation | `Optional[AuthService]` |
| `AuthService.discover() → Optional[Any]` | `Any` | `Optional[DetectionResult]` |

**8. Mixed model patterns**

Most data models use Pydantic `BaseModel` (good), but `RecoveryResult`, `CompletionResult`, and `AuthChallenge` are plain `dataclass(frozen=True)`. This means they lack `.model_dump()` and the serialization/validation consistency of the rest of the layer. `AuthChallenge.to_response()` does manual serialization as a workaround.

**9. `service.py` contains 3 classes**

`AuthStrategyRegistry`, `AuthHandle`, and `AuthService` are co-located in one 346-line file. Each has a single responsibility but they'd be easier to navigate as separate modules.

---

### Minor / cleanup

- **10.** `schemes/` folder should be renamed to `strategies/` — the port is `AuthStrategy`, the registry is `AuthStrategyRegistry`, the implementations are `OAuth2Strategy`/`ApiKeyStrategy`, but the folder says `schemes`. Leftover from old design.
- **11.** `AuthService.complete()` defaults `scheme_type` to `"oauth2"` — works today since only OAuth2 uses callbacks, but a new redirect-based scheme (SAML) would need callers to pass `scheme_type` explicitly. Consider making it required.
- **12.** Unused `import asyncio` in `mcp_provider.py`.

ui/client/src/components/agentic-ai/workspace/AuthFieldRenderer.tsx
Lina-AbuYousef: thanks <3

ui/client/src/contexts/AgenticAIContext.tsx
Lina-AbuYousef: will check this "Like btw we have the USER_ID in the metadata of the blueprint 'created_by' or something like that so we not MUST to send it." 

but why would the redis remove the need of the user id?

shared-resources/sso-backend/endpoints/__init__.py
Lina-AbuYousef: yes but did you merge it already?

shared-resources/identity/config/app_config.py
Lina-AbuYousef: like to move it to the env var?
we need it because the requests is redirected to sso and sso send them to multi agent

odaiodeh:
@Lina-AbuYousef 
These security issues needs to be handled.
 
| #  | Threat                                    | Severity | Category        |
|----|-------------------------------------------|----------|-----------------|
| 1  | Plaintext tokens in MongoDB               | CRITICAL | Data-at-rest    |
| 2  | Unauthenticated credential endpoints      | CRITICAL | AuthZ           |
| 3  | Hardcoded fallback HMAC secret            | CRITICAL | Cryptography    |
| 4  | CORS wildcard + credentials               | CRITICAL | Browser security|
| 5  | MongoDB without auth/TLS                  | HIGH     | Network         |
| 6  | Config poisoning (no tenant isolation)    | HIGH     | AuthZ           |
| 7  | client_secret echoed on save              | HIGH     | Information leak|
| 8  | Secrets in Redis flow state               | HIGH     | Data-at-rest    |
| 9  | Exception message disclosure              | HIGH     | Information leak|
| 10 | HTTP callback URL                         | MEDIUM   | Transport       |
| 11 | No rate limiting                          | MEDIUM   | Availability    |
| 12 | No credential TTL/cleanup                 | MEDIUM   | Data retention  |
| 13 | State payload leaks identity              | MEDIUM   | Privacy         |

multi-agent/lib/mas/blueprints/service.py
```diff
@@ -24,13 +24,19 @@ def __init__(
         resolver: BlueprintResolver,
         validation_service: ElementValidationService = None,
         card_service: ElementCardService = None,
+        auth_service=None,
     ):
         self._repo = repo
         self._resolver = resolver
         self._validation_service = validation_service
         self._card_service = card_service
+        self._auth_service = auth_service
```
odaiodeh, line 33: here we could be violating SRP

multi-agent/lib/mas/resources/service.py
```diff
@@ -29,18 +28,26 @@ def __init__(
...
         self._store = resource_registry
         self.element_registry = element_registry
         self._card_service = card_service
         self._dependency_resolver = DependencyResolver(resource_registry=self._store)
         self._validation_service = validation_service
+        self._auth_service = auth_service
+
+    def set_auth_service(self, auth_service) -> None:
+        """Late-bind the auth service (created after ResourcesService in the container)."""
+        self._auth_service = auth_service
```
odaiodeh, line 42: why do we couple the resource and the blueprint service to the auth service and not the validations service?

multi-agent/lib/mas/resources/service.py
odaiodeh, line 1: @Lina-AbuYousef 
`auth_service` in `ResourcesService` & `BlueprintService`

Both `ResourcesService` and `BlueprintService` accept the full `AuthService` via constructor + a late-binding `set_auth_service()` setter. I traced every usage and found several SOLID concerns.

### How `auth_service` Is Used

| Consumer | What It Does | Methods Called |
|---|---|---|
| `ResourcesService._cleanup_orphaned_credential()` | Deletes credential when no resource references that server | `delete_credential()` |
| `ResourcesService._run_pre_save_hook()` | Passes to config models' `on_pre_save()` — MCP config persists bearer tokens through it | `save_credential()` |
| `ResourcesService._validate_and_get()` | Passes into `ValidationContext` for validators | pass-through only |
| `BlueprintService._validate_spec()` | Same — passes into `ValidationContext` | pass-through only |

### Issues

**SRP Violation — `ResourcesService` owns credential lifecycle decisions**

`_cleanup_orphaned_credential()` implements a credential retention policy (orphan detection + deletion) inside a resource CRUD facade. This gives `ResourcesService` two reasons to change: resource management logic *and* credential cleanup policy. The auth domain should own the decision of when credentials are deleted, not the resource service.

**DIP Violation — Core depends on concrete `AuthService` via `Any` typing**

All injection points use untyped `Any` — `resources/service.py:31`, `blueprints/service.py:27`, `elements/common/validator.py:116`. There is no abstract interface or `Protocol` governing this dependency. The core domain (`lib/mas/resources/`, `lib/mas/blueprints/`, `lib/mas/elements/`) implicitly depends on concrete `AuthService` method names. If a method is renamed, nothing catches it at the type level.

**ISP Violation — Full `AuthService` injected, tiny slice used**

`AuthService` exposes `initiate()`, `complete()`, `detect()`, `bind()`, `bind_lazy()`, `delete_credential()`, `save_credential()`, etc. But `ResourcesService` only calls `delete_credential()` and `save_credential()`. `BlueprintService` calls **zero methods** directly — it's a pure pass-through to `ValidationContext`. Each consumer is coupled to an interface far broader than what it needs.

**Temporal Coupling — `set_auth_service()` setter**

Both services use a `set_auth_service()` setter for late-binding because they're created before `AuthService` in the container. This means the object is partially initialized between construction and the setter call. If someone forgets to call the setter, auth-dependent logic silently no-ops (`if not self._auth_service: return`) instead of failing fast. The dependency is invisible from the constructor signature alone.

### Summary

| Principle | Verdict |
|---|---|
| **SRP** | Violated — credential lifecycle logic in wrong domain |
| **OCP** | Clean — `on_pre_save(**services)` is extensible |
| **LSP** | Clean |
| **ISP** | Violated — full `AuthService` injected, 1–2 methods used |
| **DIP** | Violated — `Any` typing, no abstract interface |
| **Temporal coupling** | Smell — `set_auth_service()` partial init, fail-silent |

Not blockers, but should be addressed before this pattern gets copied to more services.

multi-agent/adapters/outbound/redis/auth_pending_store.py
sfiresht: as agreed with Lina this wil be done in phase 2 of this area.

odaiodeh (approved):

odaiodeh (approved):
i approved it for now, but i think few design changes(some SOLID violations) needs to be handled and security issues needs to be resolved.

multi-agent/lib/mas/elements/providers/mcp_server_client/config.py
```diff
@@ -22,21 +38,82 @@ class McpProviderConfig(ProviderBaseConfig):
...
+            server_identifier=server_id,
+            access_token=self.bearer_token,
+            scheme_type="api_key",
+            status=TokenStatus.ACTIVE,
+            expires_at=None,
+        ))
+
+        object.__setattr__(self, "server_identifier", server_id)
+        object.__setattr__(self, "scheme_type", "api_key")
+        object.__setattr__(self, "bearer_token", None)
```
odaiodeh, line 115: @Lina-AbuYousef what is this?

multi-agent/lib/mas/elements/providers/mcp_server_client/config.py
```diff
@@ -22,21 +38,82 @@ class McpProviderConfig(ProviderBaseConfig):
...
+        json_schema_extra=combine_hints(
+            SecretHint(allow_reveal=True),
+            ConditionalHint(visible_when={"auth_method": "access_token"}),
+        ),
     )
     additional_headers: Dict[str, Any] = Field(
         default_factory=dict,
         description="Additional HTTP headers to include in MCP server requests"
     )
+    def on_pre_save(self, user_id: str, **services) -> None:
```
odaiodeh, line 91: @Lina-AbuYousef  making logic code in config file is not correct at all.
please look at the current patterns we are doing the element, each element had spec file, it points lets say to validation file, so do you think maybe you should create life cycle file that handle specific life cycle of element?

multi-agent/lib/mas/elements/providers/mcp_server_client/mcp_provider.py
```diff
@@ -9,6 +10,11 @@
 from .provider_tool_registry import ProviderToolRegistry
 from .transport.enums import McpTransportType
 
+logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    from mas.core.auth.credentials.credential import AuthCredential
```
odaiodeh, line 16: @Lina-AbuYousef  type checking in code we don't use it, since it shows that there could be wrong direction of imports or cyclic import issues

multi-agent/lib/mas/elements/providers/mcp_server_client/mcp_provider_factory.py
```diff
@@ -1,83 +1,60 @@
...
+import logging
+from typing import Any, Dict, Optional, TYPE_CHECKING
 from mas.elements.common.base_factory import BaseFactory
 from mas.elements.common.exceptions import PluginConfigurationError
 from .config import McpProviderConfig
 from .mcp_provider import McpProvider
 from .identifiers import Identifier
 
+if TYPE_CHECKING:
+    from mas.core.auth.credentials.credential import AuthCredential
```
odaiodeh, line 12: same comment as above

multi-agent/lib/mas/elements/providers/mcp_server_client/validator.py
```diff
@@ -16,53 +23,42 @@
...
+        "clientInfo": {"name": "unifai-probe", "version": "1.0"},
+    },
+}
```
odaiodeh, line 40: @Lina-AbuYousef  i agree with maya

multi-agent/adapters/outbound/mongo/auth_token_repository.py
Lina-AbuYousef: note sure i got your question, but the cred is encrypted

multi-agent/lib/mas/actions/providers/mcp/validate_connection/validate_connection.py
Lina-AbuYousef: right need to get them back

multi-agent/lib/mas/core/auth/schemes/api_key/adapter.py
Lina-AbuYousef: this will be handled in the improving ticket

multi-agent/lib/mas/core/auth/service.py
Lina-AbuYousef: bind lazy is wrappping the auth service

multi-agent/lib/mas/elements/auths/oauth_client/oauth_client_factory.py
Lina-AbuYousef: good point

shared-resources/identity/config/app_config.py
Lina-AbuYousef: this one will be handled here this week https://redhat.atlassian.net/browse/GENIE-1565

sfiresht: as agreed will be done in  https://redhat.atlassian.net/browse/GENIE-1565

sfiresht: https://redhat.atlassian.net/browse/GENIE-1565

sfiresht (approved):
all the changes required from my side to be done in https://redhat.atlassian.net/browse/GENIE-1565

multi-agent/lib/mas/elements/providers/mcp_server_client/mcp_provider.py
Lina-AbuYousef: @odaiodeh  I'm removing them in the auth improvement ticket but please pay attention that this  type checking  is being used across multi places in the code from the part for example:
/multi-agent/lib/mas/core/contracts.py
/multi-agent/lib/mas/core/element_deps.py
and many other places, so maybe we should consider removing them also from there

multi-agent/lib/mas/elements/providers/mcp_server_client/config.py
Lina-AbuYousef: removed in auth improvement

multi-agent/lib/mas/elements/providers/mcp_server_client/config.py
Lina-AbuYousef: removed by the improvement ticket

PR #150, opened 2026-04-20
Genie 827/story/implement logout

MayaCrmi:
logout is enabled, users landing page is now:

image

## Summary by CodeRabbit

* **New Features**
  * Dedicated login page with SSO sign-in, animated UI, and auth-aware loading/redirect behavior so login is reachable outside the protected flow.
* **Bug Fixes**
  * Improved logout and token-refresh handling to clear auth state and redirect to login on failures.
  * Protected-route and 401 handling refined to avoid unwanted redirects when already on the login page.
* **Accessibility**
  * Sign-out control now uses proper button semantics and includes an ARIA label.


ui/client/src/contexts/AuthContext.tsx
MayaCrmi, line 1: AuthContext still uses /auth/user for to know who is logged in, login() redirects to the SSO, and the api still refreshes user state (or logout() if that fails). what’s new here is a clear logged-out mode: logout() hits POST /auth/logout, clears the user, sets unifai_logged_out in localStorage (and isLoggedOut in React), and goes to /login; other tabs see the flag via storage and clear auth too. after a logout the login adds prompt=login so SSO doesn’t silently re-sign you in (since it holds the token still, so now while isLoggedOut is true, startup doesn’t auto-check auth so the login page can show).

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -169,9 +170,29 @@ def auth_callback():
...
             data = self._get_server_session()
             username = (data or {}).get('user', {}).get('username', 'Unknown')
+            refresh_token_val = (data or {}).get('refresh_token') if data else None
+
+            if refresh_token_val:
+                try:
+                    keycloak_base_url = config.keycloak_base_url
+                    realm = config.get('keycloak_realm', 'master')
+                    logout_url = f"{keycloak_base_url}/realms/{realm}/protocol/openid-connect/logout"
+                    http_requests.post(
```
yhabushi79: i think you should check what is the response ( what if the server will reaponse with other code )
you treat faliures all the same
so if the server will have real issue to revoke you dont know the reason.

yhabushi79 (commented):
please see my comment.

MayaCrmi:
@yhabushi79 changed stuff both is authContext and auth_manager, take a look <3

shared-resources/sso-backend/utils/auth_manager.py
```diff
@@ -169,9 +170,38 @@ def auth_callback():
         
         @self.app.route('/api/auth/logout', methods=['POST'])
         def logout():
-            """Logout user and clear session"""
+            """Logout user and clear session (revokes refresh token at Keycloak when available)."""
             data = self._get_server_session()
             username = (data or {}).get('user', {}).get('username', 'Unknown')
+            refresh_token_val = (data or {}).get('refresh_token') if data else None
```
Nirsisr, line 176: @MayaCrmi @sfiresht saar you probable already know.
Maya till today we hold 2 variables, bearer_token & refresh token. 

Assuming the bearer valid for 10M, we should use both the bearer and the refresh to get new valid bearer with new refresh for the next cycle. I wrote a function to handle it but we decided (with omri) no need to call it. We just used the first bearer for 10H instead of taking care for those cycles to keep the token refreshed. 

With the feature @sfiresht working on plus what lina doing on other aspect, we maybe will start to keep the token valid also in GUI with levearging the function we have. FYI

Anyway the logout you did is great. Making sure we revoke the refresh_token so we can't mistakenly use it again later.

ui/client/src/components/auth/ProtectedRoute.tsx
Nirsisr, line 1: @MayaCrmi setLocation is doing the re-direction without saving the CACHE we have loaded in the GUI for CONTEXT, we get rid of it after having issue with that so I change the routing logic.

Make sure we keep this convention. Same comment I gave to yoss under the new About btn/icon, we must route in all the APP the same way without re-directions otherwise we loosing each routing the context. It's not a REDUX store like we had in TM so each re-direction affect the entire in-memory context.

ui/client/src/contexts/AuthContext.tsx
Nirsisr: @MayaCrmi I will need a short call on that, let me call you.

ui/client/src/App.tsx
Nirsisr, line 1: Get rid of `useLocation` in all the places (see me previous comment)

ui/client/src/contexts/AuthContext.tsx
```diff
@@ -136,48 +183,35 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
-  
-  // Set up token refresh and expiration checking
+  }, [isLoggedOut, checkAuthStatus]);
+
+  // Periodically refresh access token before it expires
   useEffect(() => {
-    if (!isAuthenticated || !user) return;
+    if (!isAuthenticated || !user || isLoggedOut) return;
 
     const checkTokenExpiration = () => {
```
Nirsisr, line 141: Please export the `refreshToken` to separated sub-function as where in the original code.

ui/client/src/contexts/AuthContext.tsx
Nirsisr: We had 1 on 1 call, agreed try to deal with logout functionality without saving and reference in localStorage, if we change User object to null across the entire context of the APP. Re-login should happened automatically via Maya's new login page instead of managing it via in-memory variables.

Nirsisr (changes requested):

ui/client/src/contexts/AuthContext.tsx
MayaCrmi: niros in the original code it was inside this useEffect, where it is now. do yoou want me to extract it or not?

ui/client/src/contexts/AuthContext.tsx
MayaCrmi: We didn't need the big AuthContext changes, turned it back to be as minimal changes possible, only changed what I still think is nevessary

ui/client/src/components/auth/ProtectedRoute.tsx
MayaCrmi: removed setLocation

ui/client/src/App.tsx
MayaCrmi: removed useLocation

shared-resources/sso-backend/utils/auth_manager.py
MayaCrmi: noted for the future, i don't need to change it now, correct?

Nirsisr: No need to make any change. Just put it here so you are aware for it.

Nirsisr (approved):

ui/client/src/contexts/AuthContext.tsx
Nirsisr: You return it for the original state. Perfect.

sfiresht (approved):
looks cool from my POV, approving (I assume the 15 minutes test passes).

PR #165, opened 2026-05-10
GENIE-1382-file attachment support in chat

yhabushi79:
## Summary

This PR adds the ability for users to attach files (PDF, CSV, TXT, HTML, Markdown) to their chat messages. Files are uploaded to the Gemini File API and made available to all AI agents in the multi-agent workflow via a new `read_attached_file` tool.

**Key capabilities added:**
- Paperclip button in chat UI for attaching up to 3 files (max 20MB each)
- Server-side upload to Gemini File API with retry, parallel execution, and rollback
- Automatic file reference propagation through orchestrator → delegated agents
- Agents can read/analyze attached files on demand using a dedicated tool
- File attachment chips displayed on user messages in conversation history

## Yosi's Words

users click the paperclip > attach their files ( up to 3 ) , and send. Behind the scenes, 
we upload the file to Gemini and get back file uri's exists in the workspace  The orchestrator knows files are there and explicitly tells agents to read them ( we do want to point that there are files attchmntes to read not as optonal , otherwise why to upload if it choose to ignore), then  Agents use a tool to fetch the content on-demand rather than having the full document crammed into their context window upfront.

### High-Level Flow

```
┌─────────────┐     ┌──────────────────┐     ┌─────────────────────┐
│   User UI   │────▶│  Flask Endpoint  │────▶│  Gemini File API    │
│  (attach +  │     │  (validate +     │     │  (stores file 48h)  │
│   send)     │     │   upload)        │     │                     │
└─────────────┘     └──────────────────┘     └─────────┬───────────┘
                              │                         │
                              │ file URIs               │
                              ▼                         │
                    ┌──────────────────┐                │
                    │   Graph State    │                │
                    │ (file_attachments│                │
                    │   channel)       │                │
                    └────────┬─────────┘                │
                             │                         │
                             ▼                         │
                    ┌──────────────────┐                │
                    │  Orchestrator    │                │
                    │ (sees files,     │                │
                    │  instructs       │                │
                    │  agents)         │                │
                    └────────┬─────────┘                │
                             │ delegates                │
                             ▼                         │
                    ┌──────────────────┐                │
                    │  Custom Agent    │───────────────▶│
                    │ (calls           │  read_attached │
                    │  read_attached_  │  _file tool    │
                    │  file tool)      │◀───────────────┘
                    └──────────────────┘   returns content
```

### Upload Flow (what happens when user clicks Send)

```
User selects files → Client validates (type, size, count)
       │
       ▼
FormData POST → /sessions/user.session.submit
       │
       ▼
Flask validates again (boundary guard) → reads bytes
       │
       ▼
SessionService.submit() → InputProjector.apply()
       │
       ▼
GeminiFileUploadAdapter.upload_batch()
  • Parallel upload (ThreadPoolExecutor)
  • Retry with exponential backoff (3 attempts)
  • Rollback on partial failure
       │
       ▼
FileAttachment[] stored in graph state + chat message
       │
       ▼
Workflow submitted to Temporal → HTTP 202 returned
```

### Propagation Flow (how files reach agents)

```
UserQuestionNode
  └── reads file_attachments from graph state channel
  └── stores as workspace variable: "file_attachments"
        │
        ▼
OrchestratorNode
  └── context builder reads workspace variable
  └── formats: "📎 FILE ATTACHMENTS: file.pdf → <URI>"
  └── phase guidance: "MUST instruct agents to read files"
        │
        ▼
DelegateTaskTool
  └── creates child thread
  └── copies file_attachments variable to child workspace
        │
        ▼
CustomAgentNode
  └── checks for non-expired attachments (48h TTL)
  └── injects read_attached_file tool if files exist
  └── adds "ATTACHED FILES: ..." to conversation context
  └── agent calls tool → gets file content → completes task
```


multi-agent/adapters/inbound/flask/endpoints/sessions.py
```diff
@@ -139,22 +140,78 @@ def generate():
...
 @sessions_bp.route("/user.session.submit", methods=["POST"])
 @with_require_identity_authorization
-@from_body({
-    "session_id": fields.Str(data_key="sessionId", required=True),
-    "inputs": fields.Dict(data_key="inputs", required=True),
-    "scope": fields.Str(data_key="scope", load_default="public"),
-    "session_type": fields.Str(data_key="sessionType", load_default="Personal"),
-})
-def submit_user_session(identity, session_id, inputs, scope, session_type):
+def submit_user_session(identity):
```
Lina-AbuYousef: yos pls move the functionality from the endpoint function method

multi-agent/lib/mas/core/file_attachment.py
Lina-AbuYousef, line 1: The file is at lib/mas/elements/llms/common/chat/file_attachment.py — inside the LLM chat abstraction package. But look at what actually lives in that chat/ package:

message.py — ChatMessage, Role, ToolCall
converter.py — LLM message format conversion
utils.py — chat utilities
These are all about LLM conversation modeling. file_attachment.py doesn't belong here — it's not an LLM concept. It's a session-level / domain-level concept about user-uploaded files.

Lina-AbuYousef, line 1: The file packs three distinct things:
1. The data model (FileAttachment) — a Pydantic model representing an uploaded file reference. This is a domain value object.
2. TTL/lifecycle logic (is_attachment_active, filter_active_attachments, partition_attachments) — business rules about when attachments expire. This is domain logic.
3. Prompt formatting (format_attachment_lines) — renders attachment info as text lines for LLM prompt injection. This is presentation/formatting logic specific to how agents see files.

Also get_attachment_field exists to paper over the fact that attachments travel as both FileAttachment models and plain dicts (after workspace serialization). That's a leaky abstraction — the caller shouldn't need to care, but the system forces it by serializing to dicts in workspace variables.

multi-agent/lib/mas/session/execution/ports.py
```diff
@@ -53,3 +54,62 @@ def submit(self, session: WorkflowSession, request: SubmitSessionRequest) -> Non
     def cancel(self, handle: str) -> None:
         """Request cancellation of a running background session."""
         ...
+
+
+# ── File Upload Port ──────────────────────────────────────────────────
+
+
+class FileUploadError(Exception):
```
Lina-AbuYousef: is here the right place to put the FileUpload claases?

Lina-AbuYousef (changes requested):
i need to do another cycle

PR #171, opened 2026-05-14
Genie 1584/story/ai first template

MayaCrmi:
### AI-First Template: Reporter + GitLab + Jira + Google Workspace Orchestrator
The template: orchestrator that routes user questions to four agents: a Reporter agent, a GitLab agent, a Jira agent, and a Google Workspace agent. Users instantiate it by filling in their credentials (GitLab token, Jira API token + email, Google OAuth sign-in, Gemini API key).

**What we had to add to support it:**

_Backend:_

* atlassian_user_email on McpProviderConfig — Atlassian MCP servers that use Basic-auth-style header forwarding require both an API token and the user's email as X-Atlassian-Email for the additional header. Added the field and wired it into the header-building step in the factory.
* MCP init resilience — initialization failures (401s, network errors) now degrade gracefully instead of crashing the agent, so a bad credential on one provider doesn't block the others.
* Auth field skip in instantiator — fields that carry an AuthHint (OAuth sign-in triggers) are never submitted as values, so the required-field check now skips them instead of raising a spurious "field not provided" error.
* Template draft validation in TemplateService — rejects templates that contain external $ref pointers without inline configs, preventing a class of silent runtime failures where instantiated blueprints would reference the template creator's resources instead of the new user's.

_Frontend:_

* New field types auth and validate in the template wizard — auth renders the existing AuthFieldRenderer (OAuth sign-in button) and validate renders an inline connection-check button with status feedback, matching the behaviour users already see in the inventory element form.
* anyOf $ref resolution — the schema normalisation hook now unwraps Pydantic v2's anyOf: [{$ref}, {type: null}] pattern so Optional resource models are handled correctly instead of being silently skipped.
* Conditional field visibility — fields tagged with ConditionalHint (e.g. bearer_token only when auth_method = access_token, sign_in only when auth_method = sign_in) are now respected in the template wizard, keeping the form clean.

**NOTE the json file will not be pushed - i'll add it manually to the db on the different pods. It's here for review**

## Summary by CodeRabbit

* **New Features**
  * Auth sign-in and live validation for template fields with richer status UI.
  * Validate-action support to run connection/auth checks from templates.
  * Conditional field visibility driven by template drafts and form data.
  * JSON object input rendering for object-typed fields.

* **Improvements**
  * Enhanced OAuth client registration discovery with PRM fallback and issuer remapping.
  * Provider-level streaming that preserves function-call signatures across tool use.
  * Stricter template draft validation and safer template/materialize handling.
  * Provider connection timeout handling standardized; resource-ID fields made optional.

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-community-ai-tools/UnifAI/pull/171)

### DIFF
--- a/.cursor/commands/pipeline.md
+++ b/.cursor/commands/pipeline.md
@@ -12,6 +12,11 @@ The user's input determines which mode to run. Parse the input as follows:
 ```
 To detect whether the argument is a file or a task description: check if the argument is a valid file path that exists on disk. If it is a file, read it and use it as the Phase 1 design output, then start at Phase 2. If it is not a file, treat it as a task description and start at Phase 1.
 
+**Mode 2 — design-only**: Run only Phase 1 (Design) on a task description, Jira ticket, or prompt. Stop after the design is produced. Do NOT continue to Phase 2 even if the design looks complete.
+```
+/pipeline design-only <task or Jira ticket or prompt>
+```
+
 **Mode 3 — implement**: You already have an approved design. Skip Phases 1-2. Start at Phase 3 (Implementation), using the provided file as the approved design. Continue through Phases 4-5.
 ```
 /pipeline implement <path-to-approved-design>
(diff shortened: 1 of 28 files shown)


MayaCrmi:
Closing since Lina's feature will probably make this eaiser if we ever decide this template is actually needed.

PR #181, opened 2026-05-25
Team workspace improvements

MayaCrmi:
#### 1. Fix analytics dashboard for team identities
The analytics dashboard was broken since it only tracked sessions by user_id.
I replaced the user_id field on the analytics files with identity_id, identity_type, and display_name. For this I also added aggregations for the mongoDB calls to take into account the new strucre of the field.

#### 2. Show latest activity timestamp instead of creation time in Collaboration Hub
Session cards in the Collaboration Hub (and in the regular Agentic Chats) previously showed the started_at timestamp. Now they show when the session was last interacted with as per Omri's request.
For this I added a last_active_at field and a mark_active() method. mark_finished() also sets last_active_at.

#### 3. Add carousel view toggle to Collaboration Hub
The split/chat/graph carousel mode toggle (already available in the standalone execution tab) was missing from the Collaboration Hub view.

#### 4. Fix invisible text in Share panel inputs (dark mode)
Text typed in the Share panel's search and message inputs was invisible.


multi-agent/lib/mas/core/execution_context.py
```diff
@@ -78,8 +79,12 @@ def credential_user_id(self) -> str:
             return ""
         return self.identity.id
 
+    def mark_active(self) -> ExecutionContext:
+        return self.model_copy(update={"last_active_at": datetime.now(timezone.utc)})
+
     def mark_finished(self) -> ExecutionContext:
-        return self.model_copy(update={"finished_at": datetime.now(timezone.utc)})
+        now = datetime.now(timezone.utc)
+        return self.model_copy(update={"finished_at": now, "last_active_at": now})
```
sfiresht, line 87: is this a fix or just a request from users?

MayaCrmi: Omri asked me to change the timestamp on the sessions to be last message instead of first message so I'm using the new last_active_at field

ui/client/src/components/agentic-ai/collaborationHubPanels.tsx
```diff
@@ -28,6 +28,7 @@ import type { ChatSession, ChatMessage } from "@/types/session";
 import type { FlowObject } from "./graphs/interfaces";
```
sfiresht: checking the carousel in the UI it looks like the implementation is not aligned as in the personal view it moves differently than the team mode.

ui/client/src/pages/Analytics.tsx
```diff
@@ -90,8 +90,8 @@ export default function Analytics() {
     : [];
```
sfiresht, line 90: it looks like I can't see my runs. not sure it's code related but Iput it here so we're not missing this.

ui/client/src/components/agentic-ai/collaborationHubPanels.tsx
MayaCrmi: fixed

ui/client/src/pages/Analytics.tsx
MayaCrmi: should be working uploading to PP now again

ui/client/src/hooks/use-carousel-layout.ts
```diff
@@ -0,0 +1,112 @@
...
+  containerSelector: string;
+  /** Minimum chat panel width when dragging the resizer (default 25). */
+  minPercent?: number;
+  /** Maximum chat panel width when dragging the resizer (default 80). */
+  maxPercent?: number;
+  /** When true, mode changes are blocked and mode resets to 'normal'. */
+  disabled?: boolean;
+}
+
+export function useCarouselLayout({
```
Lina-AbuYousef, line 17: so here you have extracted the code we had to a hook ?

multi-agent/adapters/outbound/mongo/session_repository.py
```diff
@@ -269,16 +277,20 @@ def get_system_analytics(
 
     # ---------- Facet Definitions ----------
 
+    _FACET_IDENTITY_TYPE_KEY = "identity_type"
```
Lina-AbuYousef: nice but why here and not in a reusable enum?

scripts/analyze_ci_failure.py
```diff
@@ -0,0 +1,486 @@
...
+    def blueprints_api_base(self) -> str:
+        """Base URL for blueprints API (different prefix)."""
+        return f"{self.base_url}{BLUEPRINTS_API_PREFIX}"
+
+    @property
+    def api_base(self) -> str:
+        return f"{self.base_url}{SESSION_API_PREFIX}"
+    
+    @property
+    def blueprints_api_base(self) -> str:
```
Lina-AbuYousef, line 113: pay attention that you have declared this twice , as well api_base

ui/client/src/hooks/use-carousel-layout.ts
MayaCrmi: yep instead of duplicating it in two places

scripts/analyze_ci_failure.py
MayaCrmi: i'm not touching this script as it's udi's team script - i just added it to our repo so we can test it when changing things to make sure it doesn't stop working (: it's from their own repository so i want it to be aligned with them

multi-agent/adapters/outbound/mongo/session_repository.py
MayaCrmi: great suggestion, i'll check where it's best to put it to be able to use it in different places

multi-agent/adapters/outbound/mongo/session_repository.py
```diff
@@ -432,5 +470,16 @@ def _determine_granularity(since: Optional[datetime], now: datetime) -> str:
...
+        fields dict so they are accessible via GroupedCount.get().
+        """
+        results = []
+        for doc in docs:
+            fields = dict(doc["_id"]) if isinstance(doc["_id"], dict) else {"_id": doc["_id"]}
+            for key, val in doc.items():
+                if key not in ("_id", "count"):
+                    fields[key] = val
+            results.append(GroupedCount(fields=fields, count=doc["count"]))
+        return results
```
odaiodeh, line 485: whats the purpose of this change what does it do?

multi-agent/lib/mas/core/execution_context.py
odaiodeh, line 87: isn't finished_at is the same as last_active_at, if so why do you need another field that has the same value?

odaiodeh (dismissed):

multi-agent/adapters/outbound/mongo/session_repository.py
MayaCrmi: its purpose it to translate from the mongo aggregation into a DTO the next layer can consume
since we changed the user id from string to object then we needed to change the aggregations

multi-agent/lib/mas/core/execution_context.py
MayaCrmi: finished_at is if the session completed and if it's None then it's still in progress, while last_ctive_at will use the latest action that happens there. so it's not the same only for runnig sessions and they serve a bit of a different needs

ui/client/src/pages/Analytics.tsx
sfiresht: doesn't work :-(

I've sent you the response from the API in slack.

MayaCrmi: seems like maybe some of the changes ruined it! pushing fixes for it now

ui/client/src/hooks/use-carousel-layout.ts
Lina-AbuYousef: nice

Lina-AbuYousef (approved):

yhabushi79 (approved):

odaiodeh (approved):

yhabushi79 (approved):

ui/client/src/pages/Analytics.tsx
sfiresht: fixed, approving

sfiresht (approved):

PR #185, opened 2026-05-27
Genie 1618/story/enhanced decorator

sfiresht:
### Maya:
Before leaving, created this branch with the goal of moving the user identification to use the session id cookie that Saar added to our Redis, and also including the team id to be taken into account at the same time.
Also, while working on it took the liberty of updating our API calls on the UI to use the method we want (http file with the endpoints, calling said functions from the different components) so a lot of the files are just that.
So now we have Login → Identity sets HttpOnly unifai_session_id cookie → Browser automatically sends cookie on every /api2 and /api3 request → MAS reads cookie.
Not that it falls back to X-Session-Id header for CLI/scripts. I haven't gotten around to it yet but we should make sure that everything is still working with these changes.

SCRIPTS that now probably won't work: execution_workflow, analyze_ci_failure.


multi-agent/adapters/inbound/flask/identity.py
```diff
@@ -0,0 +1,160 @@
...
+from mas.core.identity.ports import IdentityProvider
+
+logger = logging.getLogger(__name__)
+
+
+# ──────────────────────────────────────────────────────────────────────────────
+# Admin access gate
+# ──────────────────────────────────────────────────────────────────────────────
+
+def require_admin_access(f):
```
sfiresht, line 35: why is this here, this should be available to all modules that might require admin access.

ui/client/src/api/actions.ts
```diff
@@ -3,12 +3,40 @@ import axios from '../http/axiosAgentConfig';
...
+  });
+  return response.data;
+}
+
+export async function listActions(category: string, type: string): Promise<any[]> {
+  const response = await axios.get('/actions/actions.list', {
+    params: { category, type },
+  });
+  return response.data.actions || [];
+}
```
sfiresht, line 19: @sfiresht 
one of the tasks MAya did :

instead of bare API calls inside the code we now have only "function calls" to have all the code aligned, review this throughly :-)

Lina-AbuYousef:
As we discussed with the team, we wanted to make the decorator to retrieve the session data for user/team, now we want to make it also know to get a token from the header, and retrieve the data of the suer from the mongo and int eh end the both cases will return the same data structure of the user

Lina-AbuYousef:
so we have a db in mogno called, users, I will add a collection for tokens, that will also have the user data from the session

multi-agent/adapters/inbound/flask/identity.py
```diff
@@ -0,0 +1,160 @@
+"""
+Inbound identity adapter — Flask middleware for session validation and authorization.
+
+This module is the MAS entry point for all identity concerns at the HTTP layer:
+
+- ``with_require_team_session``: validates the Redis-backed server session and
+  resolves a user-or-team Identity.  Session ID is read from the
+  ``unifai_session_id`` HttpOnly cookie (browser) with a fallback to the
+  ``X-Session-Id`` header (CLI / scripts).
```
sfiresht, line 9: 2. Does anything actually send X-Session-Id?

No. There is no client-side code — neither in the frontend (ui/) nor the CLI (cli/) — that actually sends the X-Session-Id header. Here's what I found:

The browser frontend doesn't use it at all. It relies on the unifai_session_id cookie (sent automatically via credentials: true / supports_credentials).
The CLI (cli/unifai_cli/api/base.py) sends X-Authenticated-User header instead — not X-Session-Id:

base.py
Lines 34-38
    def _auth_headers(self, user_id: Optional[str] = None) -> dict[str, str]:
        uid = self._effective_user_id(user_id)
        if not uid:
            return {}
        return {AUTH_USER_HEADER: uid}
where AUTH_USER_HEADER = "X-Authenticated-User" (line 10).

The CORS config in flask_app.py does allow the X-Session-Id header (line 31), so the plumbing is in place on the server side.
So the X-Session-Id fallback in _get_session_id() is essentially dead code right now — it's wired up on the server to accept it, but no client ever sends it. The CLI uses a completely different auth mechanism (X-Authenticated-User), which bypasses the Redis session decorator entirely.

Lina-AbuYousef:
Add API Token Support to the Decorator
What
The require_team_session decorator (or with_require_team_session in MAS) needs a fallback path for API tokens when no Redis session is found.

Where the Token Lives
Collection: api_tokens (in the users MongoDB database)
Lookup field: token_hash (SHA-256 hash of the token)
Token prefix: unifai_t_ — if the bearer starts with this, it's an API token, not a session ID
The Document Structure
```
{
  "token_hash": "sha256 hex digest",
  "user_id": "john.doe",
  "name": "my-ci-pipeline",
  "user_data": {
    "username": "john.doe",
    "email": "john@redhat.com",
    "display_name": "John Doe",
    "sub": "keycloak-sub-id"
  },
  "created_at": "2026-06-01T...",
  "expires_at": "2026-08-30T...",
  "last_used_at": "2026-06-02T...",
  "revoked": false
}
```

The Logic to Add
In the _get_session_id function or in the decorator's validation step, before doing the Redis session lookup:
```

import hashlib
TOKEN_PREFIX = "unifai_t_"
# Check: is this an API token?
bearer = request.headers.get("Authorization", "").strip()
if bearer.startswith("Bearer "):
    token = bearer[7:].strip()
    if token.startswith(TOKEN_PREFIX):
        # Hash it and look up in MongoDB
        token_hash = hashlib.sha256(token.encode()).hexdigest()
        doc = db.api_tokens.find_one({
            "token_hash": token_hash,
            "revoked": False,
        })
        if doc and doc["expires_at"] > datetime.now(utc):
            # Valid token → resolve Identity from user_data
            user_data = doc["user_data"]
            identity = resolve_identity(user_data["username"], "user")
            # Set g.identity_username, g.identity, inject into kwargs
            # Skip Redis entirely
            return identity
        else:
            # Invalid/expired → 401
```
What the Decorator Should Do
```
1. Read bearer from Authorization header
2. If starts with "unifai_t_":
   → hash it → query MongoDB api_tokens → check not revoked + not expired
   → resolve Identity from user_data.username
   → set g.identity_username, g.identity
   → inject identity kwarg → proceed
3. Otherwise (session_id from cookie or X-Session-Id):
   → existing Redis session flow (no change)
```
What It Returns to the Endpoint
Same as the session flow — an Identity object:

`Identity(type="user", id="john.doe", display_name="John Doe")`
The endpoint never knows whether it came from a session or an API token.

Team Identity with Tokens
For now, API tokens are always user-type. If teamId is in the request body/query, the decorator should still check team membership (same as the session flow) — using user_data.username as the authenticated user.

MongoDB Access
The decorator needs access to the api_tokens collection. Options:

Pass it via container.token_repository (already wired in our AppContainer)
Or pass a validate_token callback into require_team_session
The cleanest option: add an optional token_validator: Callable[[str], Optional[UserData]] parameter to require_team_session. We supply it at startup from our TokenService.validate().

backend/api/flask/endpoints/admin_config.py
```diff
@@ -18,8 +21,11 @@
 
 
 def _get_current_user(req):
-    """Current user from X-Username/X-User-Id header (set by gateway)."""
-    return req.headers.get("X-Username") or req.headers.get("X-User-Id")
+    """Resolve current user from the unifai_session_id cookie via Redis."""
+    session_id = req.cookies.get(SESSION_COOKIE_NAME, "").strip()
```
Lina-AbuYousef, line 25: i thought this functionality will be handled in the decorator and it will only give the endpoint the user data, so no cookie stuff will be across the code

global_utils/src/global_utils/flask/decorators.py
```diff
@@ -55,50 +74,52 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
...
-    Prefer the full session decorator for API paths that need the same bar as
-    the identity service.
+                # ── 2. Resolve identity (session user + optional team) ─
+                body = request.get_json(silent=True) or {}
+
+                # teamId arrives via the natural transport for each HTTP verb:
+                #   - query param: GET requests (e.g. ?teamId=abc)
+                #   - JSON body:   POST/PUT mutations
+                #   - URL kwargs:  RESTful path params (e.g. /teams/<teamId>/...)
+                team_id = str(
```
Lina-AbuYousef, line 89: i prefer to have only one field name, not many

global_utils/src/global_utils/flask/decorators.py
```diff
@@ -55,50 +74,52 @@ def wrapped(*args: Any, **kwargs: Any) -> Any:
...
+                    request.args.get("teamId")
+                    or body.get("teamId")
+                    or kwargs.get("teamId")
+                    or kwargs.get("team_id")
+                    or ""
+                ).strip()
+
+                if team_id:
+                    if check_team_membership is not None and not check_team_membership(username, team_id):
```
Lina-AbuYousef, line 98: what is this check?

global_utils/src/global_utils/flask/decorators.py
```diff
@@ -1,52 +1,71 @@
...
+from global_utils.identity import resolve_identity
+from global_utils.redis import get_identity_session
 
 G_IDENTITY_SESSION = "identity_session"
 G_IDENTITY_USERNAME = "identity_username"
+G_IDENTITY = "identity"
 
 
-def require_identity_session(
+def require_team_session(
```
Lina-AbuYousef, line 25: why are we having this decorator here and not in identity pod

backend/api/flask/endpoints/admin_config.py
```diff
@@ -18,8 +21,11 @@
...
+    if not session_id:
+        return None
+    return get_identity_username(build_redis_client(), session_id)
```
Lina-AbuYousef, line 28: why to manage the redis client here??? pls change this

multi-agent/adapters/inbound/flask/identity.py
Lina-AbuYousef: 100% @sfiresht

multi-agent/adapters/inbound/flask/identity.py
Lina-AbuYousef, line 1: all this file should move out the multi agent

rag/infrastructure/http/session.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import Callable
+
+from flask import jsonify, request
+
+from bootstrap.app_container import redis_client
+from global_utils.redis import get_identity_username
+from global_utils.redis.constants import SESSION_COOKIE_NAME
+
+
+def _resolve_username_from_cookie() -> str | None:
```
Lina-AbuYousef, line 22: why are we doing this again and again in each component, also in rag and also in. multi agent
again, i think the right design to do is to have one place in the identity pod, its job to get a session id / useer id or idk what exactly, but the major thing is that this place is the only source of truth for retrieving user / team data from redis .. the rest of the endpoints in the app should get user/team data and thats it, no cookie access from rag

Lina-AbuYousef (changes requested):
Found one MAJOR issue (naming collision of require_admin_access in two modules with different signatures), one MAJOR integration gap (CLI still sends X-Authenticated-User header instead of X-Session-Id — all CLI authenticated calls will 401), plus several minor alignment and efficiency items.

Lina-AbuYousef:
## Architecture Proposal: Centralize Auth in the Identity Pod

### Problem

The current branch distributes auth/session logic across multiple services:

- `global_utils/flask/decorators.py` reads Redis sessions and resolves identity
- `multi-agent/adapters/inbound/flask/identity.py` wires Redis, reads cookies, checks team membership
- Every service needs direct Redis access just to know who's calling

This means if we change how sessions work, add new auth methods (API tokens), or modify team logic — we touch every service.

---

### Proposed Design

**The Identity pod should be the single owner of "who is this user?"**

All auth concerns (sessions, cookies, tokens, team membership) stay inside the Identity pod. Other services just receive the resolved identity via trusted headers set by nginx.

```
Request → Nginx → Identity Pod (/validate) → resolves identity → sets headers → Service
```

---

### Architecture Diagram

```
                         ┌─────────────────────────────────────────┐
                         │          Identity Pod                     │
                         │   (single owner of "who is this user?")  │
                         │                                           │
                         │   • SSO login (Keycloak)                  │
                         │   • Redis sessions (validate/refresh)     │
                         │   • API tokens (MongoDB hash lookup)      │
                         │   • Cookie management                     │
                         │   • Team membership (LDAP/groups)         │
                         │   • Identity resolution (user or team)    │
                         │                                           │
                         │   GET /api/auth/validate                  │
                         │   → Returns: identity headers (or 401)    │
                         └──────────────────┬──────────────────────┘
                                            │
                                            │ nginx auth_request
                                            ▼
                         ┌─────────────────────────────────────────┐
                         │              Nginx                        │
                         │                                           │
                         │   • Calls /validate on every request      │
                         │   • Sets trusted headers:                 │
                         │     X-Identity-Id                         │
                         │     X-Identity-Type                       │
                         │     X-Identity-Display-Name               │
                         │     X-Identity-Username                   │
                         │   • Forwards to backend                   │
                         └──────────────────┬──────────────────────┘
                                            │
                         ┌──────────────────┼──────────────────────┐
                         │                  │                       │
                         ▼                  ▼                       ▼
              ┌──────────────┐   ┌──────────────┐       ┌──────────────┐
              │ Multi-Agent   │   │     RAG      │       │   Backend    │
              │              │   │              │       │              │
              │ • Reads       │   │ • Reads      │       │ • Reads      │
              │   headers     │   │   headers    │       │   headers    │
              │ • Builds      │   │ • Builds     │       │ • Builds     │
              │   Identity    │   │   Identity   │       │   Identity   │
              │ • Done        │   │ • Done       │       │ • Done       │
              │              │   │              │       │              │
              │ NO Redis     │   │ NO Redis     │       │ NO Redis     │
              │ NO cookies   │   │ NO cookies   │       │ NO cookies   │
              │ NO tokens    │   │ NO tokens    │       │ NO tokens    │
              └──────────────┘   └──────────────┘       └──────────────┘
```

---

### Identity Pod: `GET /api/auth/validate`

This single endpoint handles all auth methods:

| Auth Method | Source | Lookup |
|-------------|--------|--------|
| Session cookie (`unifai_session_id`) | Browser | Redis |
| `X-Session-Id` header | CLI | Redis |
| `Authorization: Bearer unifai_t_...` | API token | MongoDB |
| Team resolution (if `teamId` present) | Any | LDAP/groups membership check |

**Returns (on 200) — headers:**

```
X-Identity-Id: john.doe (or team_id)
X-Identity-Type: user (or team)
X-Identity-Display-Name: John Doe
X-Identity-Username: john.doe (always the human)
```

**Returns 401** if invalid/expired/missing.
**Returns 403** if team membership denied.

---

### Nginx Config

```nginx
location /api/ {
    auth_request /_auth_validate;
    auth_request_set $id_type $upstream_http_x_identity_type;
    auth_request_set $id_id $upstream_http_x_identity_id;
    auth_request_set $id_display $upstream_http_x_identity_display_name;
    auth_request_set $id_username $upstream_http_x_identity_username;

    proxy_set_header X-Identity-Type $id_type;
    proxy_set_header X-Identity-Id $id_id;
    proxy_set_header X-Identity-Display-Name $id_display;
    proxy_set_header X-Identity-Username $id_username;

    proxy_pass http://backend;
}
```

---

### Service Decorator (multi-agent, RAG, backend — all the same)

```python
def with_identity(f):
    @wraps(f)
    def decorated(*args, **kwargs):
        identity_id = request.headers.get("X-Identity-Id")
        identity_type = request.headers.get("X-Identity-Type", "user")
        display_name = request.headers.get("X-Identity-Display-Name", "")

        if not identity_id:
            return jsonify({"error": "Not authenticated"}), 401

        identity = Identity(type=identity_type, id=identity_id, display_name=display_name)
        kwargs["identity"] = identity
        return f(*args, **kwargs)
    return decorated
```

**No Redis. No cookies. No token validation. No team checks.** Just read headers → build Identity.

---

### What Each Component Owns

| Concern | Owner |
|---------|-------|
| SSO login (Keycloak) | Identity pod |
| Redis sessions (create/validate/refresh) | Identity pod |
| API tokens (create/validate/revoke) | Identity pod |
| Cookie management | Identity pod |
| Team membership checks | Identity pod |
| Identity resolution (user vs team) | Identity pod |
| Passing identity to services | Nginx (trusted headers) |
| Using identity in business logic | Each service (reads headers only) |

---

### Benefits

1. **Single source of truth** — auth logic in one auditable place
2. **Services stay simple** — no Redis dependency for auth, no cookie parsing, no team lookups
3. **Easy to extend** — new auth methods only change the Identity pod
4. **Testable** — mock headers in tests, no Redis/infra needed
5. **Secure** — clients can't forge headers (nginx overwrites them)
6. **Consistent** — every service gets identity the same way

---

### Migration Path

1. Build the `/validate` endpoint on the Identity pod (handles sessions + tokens + team resolution)
2. Update nginx to call it and forward identity headers
3. Replace the complex decorator with the simple header-based one
4. Remove Redis session dependencies from multi-agent and other services

backend/api/flask/endpoints/admin_config.py
sfiresht: actually the cookie stuff is exactly what we wanted. as we get it for free and it's being secured and doesn't require code in the UI

global_utils/src/global_utils/flask/decorators.py
sfiresht: checking if the user belongs to the team sent (not sure it's needed though

global_utils/src/global_utils/flask/decorators.py
sfiresht: actually the place we wanted to put the decorator is in the global utils so it's available to all modules (all hav access to redis, mongo and other stuff so it's only logic to have them use the same code.

backend/api/flask/endpoints/admin_config.py
sfiresht: you're right, this should be moved farther up .

multi-agent/adapters/inbound/flask/identity.py
sfiresht: probably right, will try to figure out how though

rag/infrastructure/http/session.py
sfiresht: we wanted actually NOT to have the identity do all the stuff and have others call it.
we know that REDIS, mongo, vault are available to all and all modules can use the same code from global utils for repeatitive actions. (just like going to the mongo actually)

sfiresht:
closing this one as we work trhough another PR. #196.
instead of all in one go we'll separate the BEpart from the UI

PR #186, opened 2026-06-01
GENIE-1628/bug/local_dev_bug_fixes

shanitzvii:
## Changes

- Added node version detection with proper port and adapter, including user warning on mismatch
- Fixed Celery worker to use solo pool method
- Added local dev configuration for multi-agent RAG client
- Reworked services YAML parsing logic

### DIFF
--- a/local-development/LOCAL_DEV_GUIDE.md
+++ b/local-development/LOCAL_DEV_GUIDE.md
@@ -997,7 +997,7 @@ Alternatively, install Docker and the tool will auto-detect it as a fallback.
 
 #### Ghost Celery/Temporal workers after restart
 
-When you run `unifai-dev start` again (or `stop`), the tmux session is terminated via SIGHUP. Celery interprets SIGHUP as a reload signal rather than a termination signal, so the worker process can survive as an orphan. (`destroy` sends Ctrl+C first, which Celery handles correctly — but if a task is mid-flight and the 10 s timeout expires, the same SIGHUP fallback applies.) Symptoms:
+When you run `unifai-dev start` again , the tmux session is terminated via SIGHUP. Celery interprets SIGHUP as a reload signal rather than a termination signal, so the worker process can survive as an orphan. (`destroy` / 'stop' sends Ctrl+C first, which Celery handles correctly — but if a task is mid-flight and the 10 s timeout expires, the same SIGHUP fallback applies.) Symptoms:
 
 - The new worker shows a **pidbox warning**: `A node named celery@... is already using this process mailbox!`
 - Tasks dispatched from RAG are consumed by the ghost (no output in your tmux pane)
--- a/local-development/LOCAL_DEV_GUIDE.md
+++ b/local-development/LOCAL_DEV_GUIDE.md
@@ -995,6 +995,32 @@ Alternatively, install Docker and the tool will auto-detect it as a fallback.
 > [!TIP]
 > Container startup errors are captured in `/tmp/unifai-dev/logs/infra.log`. Check that file if containers silently fail to start.
 
+#### Ghost Celery/Temporal workers after restart
+
+When you run `unifai-dev start` again (or `stop`), the tmux session is terminated via SIGHUP. Celery interprets SIGHUP as a reload signal rather than a termination signal, so the worker process can survive as an orphan. (`destroy` sends Ctrl+C first, which Celery handles correctly — but if a task is mid-flight and the 10 s timeout expires, the same SIGHUP fallback applies.) Symptoms:
+
+- The new worker shows a **pidbox warning**: `A node named celery@... is already using this process mailbox!`
+- Tasks dispatched from RAG are consumed by the ghost (no output in your tmux pane)
+
+**Fix — kill the ghost manually:**
+
+```bash
+pkill -TERM -f 'celery.*worker'
+# or for temporal:
+pkill -TERM -f 'temporal-worker'
+```
+
+Then restart:
+
+```bash
+unifai-dev start rag-stack
+```
+
+> [!NOTE]
+> This only affects portless workers (celery-worker, temporal-worker). Port-based services (backend, rag, etc.) are detected and killed by the port-conflict check on every start.
+
+---
+
 #### Celery worker fails to connect
 
 If the Celery worker crashes with a connection error, RabbitMQ is likely not running. Verify:
@@ -1111,6 +1137,27 @@ Also verify you are connected to the **Red Hat SSO** — authentication-related 
 ---
 
 ### UI & Frontend
+
+#### Node.js version too old
+
+The UI requires **Node.js 22+**. If you use [nvm](https://github.com/nvm-sh/nvm) and your default version is older, `unifai-dev doctor` will report:
+
+```
+  ✖ Node.js: Node.js v18.17.0 is too old (requires 22+).
+```
+
+Fix it by switching to Node.js 22:
+
+```bash
+nvm install 22    # first time only
+nvm use 22
+```
+
+To make Node.js 22 your default so you don't have to run `nvm use` every session:
+
+```bash
+nvm alias default 22
+```
 
 #### `pnpm: command not found`
 
--- a/local-development/devtool/adapters/node_detector.py
+++ b/local-development/devtool/adapters/node_detector.py
@@ -11,7 +11,7 @@ from devtool.ports.node_resolver import NodeResolver
 class LocalNodeResolver(NodeResolver):
     """Discovers the Node.js interpreter on the local filesystem."""
 
-    def check_node(self, min_major: int) -> tuple[str, str]:
+    def check_node_js(self, min_major: int) -> tuple[str, str]:
         path = shutil.which("node")
         if not path:
             raise RuntimeError(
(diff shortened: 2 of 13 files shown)


multi-agent/lib/mas/elements/providers/rag_client/config.py
```diff
@@ -1,9 +1,12 @@
+import os
 from typing import Literal
 from pydantic import Field, HttpUrl
 from mas.elements.providers.common.base_config import ProviderBaseConfig
 from mas.core.field_hints import ActionHint, HintType
 from .identifiers import Identifier
 
+_DEFAULT_RAG_URL = os.environ.get("RAG_BASE_URL", "http://unifai-rag-server:13456")
+
```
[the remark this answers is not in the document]
shanitzvii: Unlike other providers, the RAG provider is internal — its spec is not
    registered in the ElementRegistry and it never appears in the UI catalog
    or goes through the blueprint build pipeline.  It is only consumed in
    code by DocsRagRetriever, its validator, and the RAG discovery actions.

Because of this, and because we wanted to keep design properties and not do OS operations, I made inherits RagProviderConfig from SharedConfig (BaseSettings) instead of ProviderBaseConfig (BaseModel), so all fields are automatically overridable via env vars prefixed with ``RAG_`` (e.g. ``RAG_BASE_URL``),``.env`` files, YAML, or JSON sources.

    
 I also  cleaned up the dead code: removed the type discriminator field (not in the registry pipeline), the ActionHint (no UI form exists), and the unused imports that came with them.
   
   please note that when you review @Nirsisr @odaiodeh

shanitzvii:
Unlike other providers, the RAG provider is internal — its spec is not
registered in the ElementRegistry and it never appears in the UI catalog
or goes through the blueprint build pipeline. It is only consumed in
code by DocsRagRetriever, its validator, and the RAG discovery actions.

Because of this, and because we wanted to keep design properties and not do OS operations, I made inherits RagProviderConfig from SharedConfig (BaseSettings) instead of ProviderBaseConfig (BaseModel), so all fields are automatically overridable via env vars prefixed with RAG_ (e.g. RAG_BASE_URL),.env files, YAML, or JSON sources.

I also cleaned up the dead code: removed the type discriminator field (not in the registry pipeline), the ActionHint (no UI form exists), and the unused imports that came with them.

please note that when you review @Nirsisr @odaiodeh

local-development/services.yaml
```diff
@@ -71,6 +74,7 @@ services:
     infrastructure: [mongo, redis, temporal]
     env_file: ".env"
     env_entries:
+      RAG_BASE_URL: "http://127.0.0.1:13457"
```
Lina-AbuYousef: are we okay with adding a local host ip here? with fixed port

local-development/devtool/adapters/node_detector.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            out = subprocess.check_output(
+                [path, "--version"], text=True, timeout=10,
+            ).strip()
+        except (subprocess.CalledProcessError, FileNotFoundError, subprocess.TimeoutExpired) as e:
+            raise RuntimeError(
+                f"Found node at {path} but could not determine its version."
+            ) from e
+
+        ver_str = out.lstrip("v")
```
Lina-AbuYousef, line 31: what is this strip solving?

local-development/devtool/adapters/node_detector.py
```diff
@@ -0,0 +1,46 @@
...
+import shutil
+import subprocess
+
+from devtool.ports.node_resolver import NodeResolver
+
+
+class LocalNodeResolver(NodeResolver):
+    """Discovers the Node.js interpreter on the local filesystem."""
+
+    def check_node(self, min_major: int) -> tuple[str, str]:
```
Nirsisr: @shanitzvii Maybe just change the method name to check_node_js (since we can have ambiguity with the word node in Unifai : d)

local-development/devtool/adapters/registry_loader.py
```diff
@@ -111,6 +117,23 @@ def _parse_groups(raw: dict) -> dict[str, ServiceGroup]:
             for name, svc_list in raw.items()
         }
 
+    @staticmethod
+    def _parse_node_min(raw: dict) -> int | None:
```
Nirsisr: Same here Shani, for clarity.

local-development/LOCAL_DEV_GUIDE.md
```diff
@@ -995,6 +995,32 @@ Alternatively, install Docker and the tool will auto-detect it as a fallback.
...
+
+```bash
+unifai-dev start rag-stack
+```
+
+> [!NOTE]
+> This only affects portless workers (celery-worker, temporal-worker). Port-based services (backend, rag, etc.) are detected and killed by the port-conflict check on every start.
+
+---
+
```
Nirsisr, line 1023: @shanitzvii We hope to get rid of it, in the next CR cycle for local-dev.

Once we are moving both Temopral & Celery workers live under dedicated pods, I guess those comment won't be relevant anymore. So for next cycle just remember to remove it.

multi-agent/lib/mas/elements/providers/rag_client/config.py
```diff
@@ -1,25 +1,21 @@
...
-    Configuration for RAG service client.
-    Connects to a RAG server for vector database queries and document retrieval.
+    Infrastructure config for RAG service client.
+
+    Unlike other providers, RAG is internal (not in the ElementRegistry or
+    UI catalog), so it inherits SharedConfig instead of ProviderBaseConfig —
+    fields are overridable via ``RAG_``-prefixed env vars / ``.env``.
     """
-    type: Literal[Identifier.TYPE] = Identifier.TYPE
+    model_config = SettingsConfigDict(env_prefix="RAG_")
```
Nirsisr: @shanitzvii 1 on 1 on this one for 2 min, maybe you confused or my-self. (wanna make sure)

SharedConfig as part of how it works will take ENV vars in priority (it means that if I declare ENV var on my running env/VM it will use it in each class inherit from SharedConfig).

Why do we need this too? We have .env you declare also env vars under, so it will already consume them since it inherits from SharedConfig.

Nirsisr (changes requested):
Great work with Hexagonal architecture, keep pushing.

Nirsisr:
> Unlike other providers, the RAG provider is internal — its spec is not registered in the ElementRegistry and it never appears in the UI catalog or goes through the blueprint build pipeline. It is only consumed in code by DocsRagRetriever, its validator, and the RAG discovery actions.
> 
> Because of this, and because we wanted to keep design properties and not do OS operations, I made inherits RagProviderConfig from SharedConfig (BaseSettings) instead of ProviderBaseConfig (BaseModel), so all fields are automatically overridable via env vars prefixed with RAG_ (e.g. RAG_BASE_URL),.env files, YAML, or JSON sources.
> 
> I also cleaned up the dead code: removed the type discriminator field (not in the registry pipeline), the ActionHint (no UI form exists), and the unused imports that came with them.
> 
> please note that when you review @Nirsisr @odaiodeh

I put a comment on that specific change. I think the inheritance you added is completely right. 

Regard the prefixed with RAG_ on .env files not sure I understand it. If we have the .env files in place so those ENV VARS configured in the VM which makes SharedConfig use it. (why do we need this RAG_ too.

multi-agent/lib/mas/elements/providers/rag_client/config.py
shanitzvii: you're right in general, the reason I added this line is because without the prefix, it would look for BASE_URL, TOP_K, TIMEOUT as env vars, which are too generic, easy to collide with other services. The RAG_ prefix narrow the scopeand  namespaces them so only RAG_BASE_URL  picked up.

local-development/services.yaml
shanitzvii: yes because it's injected environment variable which saved on local .env files (that created during the local-dev init process, per user)
it's easy for user to change it and not affecting the deployment :)

local-development/devtool/adapters/node_detector.py
shanitzvii: removes any trailing newline (and leading/trailing whitespace) from the output of node --version. :)

local-development/LOCAL_DEV_GUIDE.md
shanitzvii: np, as discussed 1:1 :) 
for next improvements PR

local-development/devtool/adapters/node_detector.py
shanitzvii: done :)

local-development/devtool/adapters/registry_loader.py
shanitzvii: done :)

multi-agent/lib/mas/elements/providers/rag_client/config.py
shanitzvii: after 1:1 - removed :)

Lina-AbuYousef (approved):

Nirsisr (approved):

PR #193, opened 2026-06-08
fixes

Lina-AbuYousef:
- Sets `socket_timeout=30` on all Redis connection pools in multi-agent to fix session streaming hang.
- **Root cause**: `redis-py` 8.0.0 changed the default `socket_timeout` from `None` to `5s`. This causes `XREAD BLOCK 5000` to collide with the socket timeout, making the read fail/hang instead of returning an empty result after 5 seconds.
- This broke `session.subscribe` heartbeats — during agent "thinking" periods (>5s), no keepalive packets were sent, causing sessions to hang indefinitely.
## Affected files
- `multi-agent/adapters/outbound/channels/redis/factory.py`
- `multi-agent/bootstrap/container.py`
- `multi-agent/adapters/outbound/redis/collaboration_store.py`

### DIFF
--- a/multi-agent/bootstrap/container.py
+++ b/multi-agent/bootstrap/container.py
@@ -150,7 +150,7 @@ class AppContainer(metaclass=SingletonMeta):
         pending_store = None
         if redis_url:
             import redis as redis_lib
-            redis_client = redis_lib.Redis.from_url(redis_url)
+            redis_client = redis_lib.Redis.from_url(redis_url, socket_timeout=30)
             pending_store = RedisFlowStateStore(
                 redis_client=redis_client,
                 encryption_key=cfg.credential_encryption_key,
--- a/multi-agent/adapters/outbound/channels/redis/factory.py
+++ b/multi-agent/adapters/outbound/channels/redis/factory.py
@@ -28,7 +28,7 @@ class RedisChannelFactory(ChannelFactory):
         block_ms: int = 5000,
         batch_size: int = 50,
     ) -> None:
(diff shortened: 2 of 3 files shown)


Nirsisr (approved):

Nirsisr:
Amazing work Lina, 10x for that.

@odaiodeh @sfiresht please approve this PR ASAP, this is in highest priority merge it to main.

multi-agent/bootstrap/container.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, cfg: AppConfig):
         pending_store = None
         if redis_url:
             import redis as redis_lib
-            redis_client = redis_lib.Redis.from_url(redis_url)
+            redis_client = redis_lib.Redis.from_url(redis_url, socket_timeout=30)
```
odaiodeh, line 153: just not relevant question to the PR, why are you initing redis_client in the contianer, why not sending the reddis url to the RedisFlowStateStore, and let it create the client? also why are you we getting the reddis url twice in app container

Lina-AbuYousef: i dont think this is a good approach , I think the better approach is to init the redis client only once in the container and to send it to any other instance that needs it.. 
regarding " also why are you we getting the reddis url twice in app container" I can happely take care of it in another branch, just to mentioned this is not related to this or the auth branch :p , multiple different branches did this approach , i agree on this one

odaiodeh (approved):

PR #200, opened 2026-06-15
feat(backend): Add Slack slash commands integration

Lina-AbuYousef:
Implements a full `/unifai` slash command system following hexagonal architecture with proper separation of concerns.

## Architecture

```
backend/slack_commands/
├── clients/multiagent.py      — Outbound adapter (all HTTP to multi-agent)
├── execution/session_executor.py — Async session lifecycle (poll/respond)
├── commands/                  — Thin command handlers (one per subcommand)
├── models.py                  — Domain models (SlackCommand, SlackResponse)
├── formatters.py              — Presentation layer (Slack message formatting)
└── service.py                 — Command router (dispatch by subcommand)
```

## Supported Commands

- `/unifai help` — Show available commands
- `/unifai health` — Check service status
- `/unifai whoami` — Show Slack identity info
- `/unifai blueprints` — List available blueprints
- `/unifai list [page]` — List user sessions (paginated)
- `/unifai ask <blueprint|session_id> <question>` — Run or continue a session
- `/unifai status <session_id>` — Check session status and metadata
- `/unifai history <session_id>` — View session chat history
- `/unifai cancel <session_id>` — Cancel a running session
- `/unifai delete <session_id>` — Permanently delete a session


backend/entrypoint.sh
```diff
@@ -14,14 +14,20 @@ case "$ROLE" in
     exec gunicorn -w $GUNICORN_WORKERS --threads $GUNICORN_THREADS -b 0.0.0.0:$PORT --timeout $GUNICORN_TIMEOUT --access-logfile - --error-logfile - run.wsgi:application
     ;;
 
+  slack-socket)
```
sfiresht, line 17: I'm a bit confused here, is this a separate role than the regular BE? if the answer is yes than I'd expect to still see here the "flask" option (which I don't)
I assume the answer is no and the BE works regularly including a web socket, in that case, any reason to rename the BE role to a very specific one which can be confusing?
the reason I ask is that the helm now includes a deployment for which the role is always slack -socket. so the BE won't work properly it seems.

backend/slack_commands/socket_handler.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+
+from slack_bolt import App
+from slack_bolt.adapter.socket_mode import SocketModeHandler
+
+from config.app_config import AppConfig
+from core.app_container import AppContainer
+from slack_commands.models import SlackCommand
+
+logging.basicConfig(
```
sfiresht: Isn't the logging being "created" by some shared folder or something so you can just call the get_logger function?

helm/backend/unifai-backend/templates/deployment.yaml
```diff
@@ -73,6 +73,23 @@ spec:
           volumeMounts:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+        {{- if .Values.slackSocket.enabled }}
+        - name: slack-socket
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+          - name: ROLE
```
sfiresht, line 81: nay reason we need a separate one? 
anyway, I didn't see any elaboration on the deployment part in the PR description so I try to fill in from my own experience, can you explain a bit on that?

backend/slack_commands/commands/list_blueprints.py
```diff
@@ -0,0 +1,31 @@
...
+
+from slack_commands.clients.multiagent import MultiagentClient
+from slack_commands.commands.base import CommandHandler
+from slack_commands.formatters import format_blueprint_list
+from slack_commands.models import SlackCommand, SlackResponse
+
+logger = logging.getLogger(__name__)
+
+
+class ListBlueprintsCommand(CommandHandler):
```
sfiresht, line 10: arch question :-)
I dunno if we plan to extend the scope of the commands bu if that to be the case won't it better to have it split like 
FOLDERS:
list
create
run

and in each: blueprints, workflows node etc.?

just being curious .

backend/slack_commands/commands/history.py
```diff
@@ -0,0 +1,85 @@
...
+from slack_commands.clients.multiagent import MultiagentClient
+from slack_commands.commands.base import CommandHandler
+from slack_commands.models import SlackCommand, SlackResponse, sanitize_slack_arg
+
+logger = logging.getLogger(__name__)
+
+_MAX_MESSAGES = 10
+_MAX_CONTENT_LENGTH = 300
+
+_ROLE_EMOJI = {
```
sfiresht: as for the constants here and in the other EMOJIES is it preferable to have them as some external constants file and just import them here?

backend/slack_commands/clients/multiagent.py
```diff
@@ -0,0 +1,173 @@
...
+        resp = requests.get(
+            f"{self._base_url}/api/sessions/session.meta",
+            params={"sessionId": session_id},
+            headers=self._auth_headers(user_name),
+            timeout=self._timeout,
+        )
+        resp.raise_for_status()
+        return resp.json()
+
+    def _auth_headers(self, user_name: str) -> dict:
```
sfiresht: we need to find a solution for this as this should be removed as it creates a sec. breach, how do we identity user between internal modules?
I have the same "issue" in the file separation between MA and RAG - I used a different approach of passing through the cookie from the UI in that case (in your case this should be somthing else since it's slack) as eventually we want to remove this header option.

global_utils/src/global_utils/flask/decorators.py
```diff
@@ -268,3 +274,69 @@ def decorated_function(*args, **kwargs):
...
+    Returns 401 on failure.
+    """
+    def decorator(f: Callable) -> Callable:
+        @wraps(f)
+        def wrapped(*args: Any, **kwargs: Any) -> Any:
+            signing_secret = get_signing_secret()
+
+            if not signing_secret:
+                return f(*args, **kwargs)
+
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/execution/session_executor.py
```diff
@@ -0,0 +1,164 @@
...
+    def _post_to_slack(response_url: str, text: str, success: bool = False) -> None:
+        try:
+            requests.post(
+                response_url,
+                json={
+                    "response_type": "in_channel" if success else "ephemeral",
+                    "text": text,
+                },
+                timeout=10,
+            )
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/service.py
```diff
@@ -0,0 +1,45 @@
...
+                ),
+            )
+
+        try:
+            return handler.handle(command)
+        except Exception as e:
+            logger.error(
+                "Command '%s' failed: %s", command.subcommand, e, exc_info=True
+            )
+            return SlackResponse(text=f":x: Command failed: {e}")
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/clients/multiagent.py
```diff
@@ -0,0 +1,173 @@
...
+        try:
+            resp = requests.get(
+                f"{self._base_url}/api/sessions/session.status.get",
+                params={"sessionId": session_id},
+                headers=self._auth_headers(user_name),
+                timeout=5,
+            )
+            return resp.status_code == 200
+        except Exception:
+            return False
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/commands/cancel.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            cancelled = self._client.cancel_session(session_id, command.user_name)
+        except requests.HTTPError as e:
+            if e.response is not None and e.response.status_code == 404:
+                return SlackResponse(text=f":x: Session `{session_id}` not found.")
+            logger.error("Cancel failed: %s", e, exc_info=True)
+            return SlackResponse(text=f":x: Failed to cancel session: {e}")
+        except requests.Timeout:
+            return SlackResponse(text=":hourglass: Multi-agent service timed out.")
+        except Exception as e:
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/entrypoint.sh
Lina-AbuYousef: fixed (ROLE: "flask" added to Helm values)

backend/slack_commands/socket_handler.py
Lina-AbuYousef: fixed (uses shared.logger) , thanks for the great feedback

backend/slack_commands/commands/history.py
Lina-AbuYousef: moved _ROLE_EMOJI to formatters.py

backend/slack_commands/clients/multiagent.py
Lina-AbuYousef: fixed (HMAC signing implemented),  The multiagent.py client file is already deleted, and all requests now use cryptographic HMAC signatures instead of the raw header.

helm/backend/unifai-backend/templates/deployment.yaml
Lina-AbuYousef: yes, but what is the question exactly?

backend/slack_commands/commands/list_blueprints.py
Lina-AbuYousef: I considered it but with only ~8 commands the flat structure is simpler to navigate. If the scope grows significantly (multiple verbs × resources), we'd restructure by resource rather than by verb to keep related logic together.

backend/slack_commands/service.py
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/api/flask/endpoints/slack_commands.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Receive a Slack slash command POST (application/x-www-form-urlencoded).
+
+    Parses into a SlackCommand, delegates to SlackCommandsService,
+    and returns the formatted Slack JSON response.
+    """
+    command = SlackCommand.from_form(request.form)
+    service = current_app.container.slack_commands_service
+    response = service.execute(command)
+    return jsonify(response.to_dict()), 200
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/config/app_config.py
```diff
@@ -9,4 +9,6 @@ class AppConfig(SharedConfig):
     version: str = "1.0.0"
     admin_allowed_users: list = []
     rag_url: str = "http://localhost:13457"
+    multiagent_url: str = "http://localhost:8002"
+    slack_signing_secret: str = ""
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/clients/multiagent.py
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/commands/ask.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def handle(self, command: SlackCommand) -> SlackResponse:
+        parts = command.args.split(maxsplit=1)
+
+        if len(parts) < 2:
+            return self._usage()
+
+        ref, question = sanitize_slack_arg(parts[0]), parts[1]
+
+        if _UUID_PATTERN.match(ref) and self._client.session_exists(ref, command.user_name):
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/commands/history.py
```diff
@@ -0,0 +1,85 @@
...
+            content = msg.get("content") or msg.get("text") or ""
+            emoji = _ROLE_EMOJI.get(role.lower(), ":speech_balloon:")
+
+            if isinstance(content, list):
+                content = " ".join(
+                    part.get("text", "") for part in content if isinstance(part, dict)
+                )
+
+            if len(content) > _MAX_CONTENT_LENGTH:
+                content = content[:_MAX_CONTENT_LENGTH] + "…"
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/commands/status.py
```diff
@@ -0,0 +1,65 @@
...
+        except requests.HTTPError as e:
+            if e.response is not None and e.response.status_code == 404:
+                return SlackResponse(text=f":x: Session `{session_id}` not found.")
+            logger.error("Status check failed: %s", e, exc_info=True)
+            return SlackResponse(text=f":x: Failed to get status: {e}")
+        except requests.Timeout:
+            return SlackResponse(text=":hourglass: Multi-agent service timed out.")
+        except Exception as e:
+            logger.error("Status check failed: %s", e, exc_info=True)
+            return SlackResponse(text=f":x: Unexpected error: {e}")
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/execution/session_executor.py
```diff
@@ -0,0 +1,164 @@
...
+                )
+
+        except requests.HTTPError as e:
+            status_code = e.response.status_code if e.response is not None else "?"
+            body = self._extract_error_body(e)
+            logger.error("Session HTTP error: %s %s", status_code, body, exc_info=True)
+            self._post_to_slack(
+                response_url,
+                f":x: Request failed (HTTP {status_code}): {body or e}",
+            )
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/execution/session_executor.py
```diff
@@ -0,0 +1,164 @@
...
+            or result.get("answer")
+            or result.get("result")
+        )
+
+        if not answer and isinstance(result, dict):
+            messages = result.get("messages", [])
+            if messages and isinstance(messages, list):
+                last = messages[-1]
+                answer = last.get("content") or last.get("text") or str(last)
+
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/execution/session_executor.py
```diff
@@ -0,0 +1,164 @@
...
+            requests.post(
+                response_url,
+                json={
+                    "response_type": "in_channel" if success else "ephemeral",
+                    "text": text,
+                },
+                timeout=10,
+            )
+        except Exception as e:
+            logger.error("Failed to post deferred response to Slack: %s", e)
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/formatters.py
```diff
@@ -0,0 +1,88 @@
...
+        title = (
+            session.get("title")
+            or session.get("metadata", {}).get("title")
+            or ""
+        )
+
+        emoji = _STATUS_EMOJI.get(status.upper(), ":grey_question:")
+        label = title if title else blueprint_id or "untitled"
+
+        lines.append(f"{emoji} `{session_id}` — {label} ({status})")
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/formatters.py
```diff
@@ -0,0 +1,88 @@
...
+    if not sessions:
+        return SlackResponse(text=":inbox_tray: You have no sessions yet.")
+
+    total = len(sessions)
+    total_pages = math.ceil(total / page_size)
+    page = min(page, total_pages)
+
+    start = (page - 1) * page_size
+    end = start + page_size
+    page_sessions = sessions[start:end]
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/commands/delete.py
```diff
@@ -0,0 +1,46 @@
...
+                text="*Usage:* `/unifai delete <session_id>`"
+            )
+
+        try:
+            resp = requests.delete(
+                f"{self._url}/api/sessions/session.delete",
+                params={"sessionId": session_id},
+                headers=auth_headers(self._secret, command.user_id),
+                timeout=MAS_TIMEOUT,
+            )
```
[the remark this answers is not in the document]
Lina-AbuYousef: Acknowledged — this is a pre-existing gap in the MAS endpoint definitions, not introduced by this PR. Tracked for a follow-up to add auth decorators to session.cancel and session.delete routes

backend/slack_commands/commands/list_blueprints.py
Nirsisr: I agree with Saar here, if we have permanent command such as LIST per different resource we might go with that structure.

But from the other hand most of the COMMANDS we have for `sessions` are not the same as for `blueprints` and vice-versa so it's might be not right. In CLI I did sessions/ blutprints/ and under the relevant folders all the API supported for those layers.

backend/slack_commands/commands/session_status.py
Nirsisr, line 1: @Lina-AbuYousef The name here is a bit misleading,I think it should be `session_status`

backend/slack_commands/formatters.py
Nirsisr, line 1: @Lina-AbuYousef maybe the formatter for each list command should be an abstracted method which can be implemented or not (not mandatory) under the LIST sub-folders.

E.G `format_session_list` maybe should be in the command level folder.

backend/slack_commands/models.py
Nirsisr, line 1: @Lina-AbuYousef Follow my prev comment, maybe here it's the right place for formatter under `SlackCommand` abstract method.

Nirsisr (changes requested):
@Lina-AbuYousef Looking great, putting few comments.

I would like to get 5 min call explanation on `HMAC` (I feel it's the only missing part here for me, I want to understand how does it work)

We should have another ticket in the future covering unifying between `commands` folder to CLI sub-folder containing commands (since both are almost the same)

backend/config/app_config.py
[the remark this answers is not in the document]
Lina-AbuYousef: this values are being taken from env var

ci/pipeline-deploy.groovy
```diff
@@ -42,7 +42,7 @@ def secret_lists = [
     rabbitmq: ['rmq_username', 'rmq_password'],
     umami: ['umami_username', 'umami_password'],
     // keycloak: ['keycloak_base_url', 'client_id', 'client_secret', 'keycloak_realm'],
-    global_config: ['secret_key', 'vault_role_id', 'vault_secret_id', 'langfuse_base_url', 'langfuse_public_key', 'langfuse_secret_key'],
+    global_config: ['secret_key', 'vault_role_id', 'vault_secret_id', 'langfuse_base_url', 'langfuse_public_key', 'langfuse_secret_key', 'slack_signing_secret', 'slack_app_token', 'slack_bot_token'],
```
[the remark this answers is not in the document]
Lina-AbuYousef: we removed SERVICE_SIGNING_SECRET, check it again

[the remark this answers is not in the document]
Lina-AbuYousef: its being taken from vault

backend/slack_commands/commands/list_blueprints.py
Lina-AbuYousef: good point! I agree we should align the Slack commands structure with the CLI folder layout in a follow-up ticket. For now the flat structure works since we only have a handful of commands, but as we add more it makes sense to group by resource (sessions/, blueprints/)

backend/slack_commands/commands/session_status.py
Lina-AbuYousef: agree, fixed

backend/slack_commands/models.py
Lina-AbuYousef: I'd prefer keeping SlackCommand as a pure data model and the formatting logic in the handlers/formatters. Mixing formatting into the model would couple presentation with data. But when we do the folder restructuring, each command can own its formatter more naturally.

backend/slack_commands/formatters.py
Lina-AbuYousef: I see the point about co-locating formatters with their commands. Right now the shared formatters.py keeps things DRY since STATUS_EMOJI and ROLE_EMOJI are used across multiple handlers. When we restructure into resource folders , we can move the formatters into their respective folders

backend/slack_commands/models.py
```diff
@@ -0,0 +1,87 @@
...
+            if not form.get(field):
+                return field
+        return None
+
+    @classmethod
+    def from_form(cls, form: dict) -> "SlackCommand":
+        raw_text = (form.get("text") or "").strip()
+        public = bool(_PUBLIC_FLAG.search(raw_text))
+        text = _PUBLIC_FLAG.sub("", raw_text).strip()
+
```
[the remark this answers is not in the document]
Lina-AbuYousef: fixed

backend/slack_commands/commands/list_blueprints.py
sfiresht: ok, agreed, let's have a story for that, feel free to resolve this thread after story is created.

backend/slack_commands/clients/multiagent.py
sfiresht: please tell me we didn't implement hmac yet... and leave it as is with the x authenticated user header

helm/backend/unifai-backend/templates/deployment.yaml
sfiresht: a bit on the web socket or /and on the design from deployment POV, for example, why do we need s eparate pod just for the websocket :-)

backend/entrypoint.sh
sfiresht: as agreed add this to imprvemments story:
1. do we need a real seaprate container?
2.aligne the deployment file to use params and not hardcoded values

Lina-AbuYousef: https://redhat.atlassian.net/browse/GENIE-1723 follow up ticket

backend/slack_commands/commands/list_blueprints.py
sfiresht: resolving as we got an improvement story

sfiresht (approved):

Nirsisr (approved):

PR #204, opened 2026-06-15
Bugfix/deep agent

odaiodeh:


multi-agent/lib/mas/elements/llms/common/langchain_adapter.py
```diff
@@ -94,15 +94,25 @@ def _stream(
...
                     message=LangChainConverter.to_lc_message_chunk(token_or_msg),
                 )
+                yielded = True
+        if not yielded:
+            # LangChain raises "No generations found in stream" when _stream()
+            # yields zero chunks.  Emit a minimal empty chunk so the downstream
+            # generate_from_stream() always receives at least one generation.
+            yield ChatGenerationChunk(
+                message=AIMessageChunk(content=""),
+            )
```
Nirsisr, line 115: How this could happen in practice? Someone configured `Deep Agent` with one of our LLMs and it yield even no 1 token? 

It doesn't look reasonable for me. If we have the `heartbeat` in place it will guarantee the continuation of the stream, why do we need to yield Chunks from the LLM? Like in what case we see the LLM generate 0 tokens at all?

Nirsisr (changes requested):
@odaiodeh I am not sure about this ticket, please see my comment. Maybe worth to investigate this issue a bit more before taking this approach.

multi-agent/lib/mas/elements/llms/common/langchain_adapter.py
```diff
@@ -94,15 +94,25 @@ def _stream(
...
                 )
+                yielded = True
             elif isinstance(token_or_msg, ChatMessage):
                 yield ChatGenerationChunk(
                     message=LangChainConverter.to_lc_message_chunk(token_or_msg),
                 )
+                yielded = True
+        if not yielded:
+            # LangChain raises "No generations found in stream" when _stream()
+            # yields zero chunks.  Emit a minimal empty chunk so the downstream
```
Nirsisr, line 111: Like this comment feels to me like not reasonable behavior, how it even can happen?

PR #234, opened 2026-07-09
fix(slack): allow UUID to resolve as workflow ID when not a session

Lina-AbuYousef:
When a UUID doesn't match an existing session, fall through to _resolve_workflow instead of returning "Session not found". This lets users start new sessions with a workflow UUID directly.

### DIFF
--- a/backend/slack_commands/commands/ask.py
+++ b/backend/slack_commands/commands/ask.py
@@ -41,20 +41,18 @@ class AskCommand(CommandHandler):
                 return SlackResponse(
                     text=":x: Could not verify session. Please try again.",
                 )
-            if not exists:
-                return SlackResponse(
-                    text=f":x: Session `{ref}` not found.",
+            if exists:
+                self._executor.continue_session(
+                    user_name=command.user_name,
+                    session_id=ref,
+                    question=question,
+                    response_url=command.response_url,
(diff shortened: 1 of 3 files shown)


backend/slack_commands/commands/ask.py
```diff
@@ -36,25 +36,18 @@ def handle(self, command: SlackCommand) -> SlackResponse:
         ref, question = sanitize_slack_arg(parts[0]), parts[1]
 
         if _UUID_PATTERN.match(ref):
-            exists = self._session_exists(ref, command.user_name)
-            if exists is None:
```
MayaCrmi, line 40: we don't need this check anymore?

MayaCrmi (approved):

backend/slack_commands/commands/ask.py
Lina-AbuYousef: Yes, The exists is None path blocked UUIDs from ever being tried as workflow IDs, when MAS returns 500 (instead of 404) for non-existent sessions, the user was stuck with "Could not verify session". Now the code simply falls through: if the session check fails for any reason, the UUID is treated as a workflow ID. If MAS is truly down, the workflow creation will also fail with a clear error anyway.

PR #235, opened 2026-07-09
fix(identity): preserve team names in Redis cache to fix analytics di…

yhabushi79:
## Root cause

When a team workflow runs, the system needs the team name. It asks Redis for it, but Redis only stored team IDs — not names. So it returned the UUID as the name, and that UUID got saved permanently in MongoDB.

**Before (broken):**

```json
["1252c5bd...", "03f76cd6...", "af6b942d..."]
```

**After (fixed):**

```json
[
  {"team_id": "1252c5bd...", "name": "123"},
  {"team_id": "03f76cd6...", "name": "ccxas"},
  {"team_id": "af6b942d...", "name": "test-team"}
]
```

## The fix

Changed Redis to store team names alongside IDs. Now when the system asks "what's this team's name?", Redis has the answer.

**Files changed:**

- `global_utils/src/global_utils/redis/team_cache.py` — store/read `[{team_id, name}]` instead of `[id]`
- `global_utils/src/global_utils/identity_client.py` — preserve team name when writing to cache

## The migration script

Old sessions in MongoDB already have the UUID saved as the team name. The script finds those records, looks up the real name from the Identity database, and patches them.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,7 +6,7 @@ backend/ @sfiresht @yhabushi79 @odaiodeh @nirsisr @Lina-AbuYousef @MayaCrmi
 rag/ @odaiodeh @nirsisr @Lina-AbuYousef
 ui/ @nirsisr @Lina-AbuYousef @MayaCrmi
 shared-resources/ @yhabushi79 @nirsisr @sfiresht
-multi-agent/ @odaiodeh @nirsisr @Lina-AbuYousef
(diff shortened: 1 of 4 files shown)


MayaCrmi (dismissed):
Code looks good, I'm not sure if this is the only place we'd see this change on the UI so just make sure nothing else under Team etc is changed :D

Nirsisr (approved):

Nirsisr:
Followed our call, I understand why the change where required (GUI visualization improvement under Analytics).

Code changes looks smooth, approved.

PR #241, opened 2026-07-21
Genie 1648/story/scheduled workflow prompts

shanitzvii:
**Scheduled Workflows**

Adds scheduled workflow execution: users can configure recurring (cron/interval) or one-off runs from the UI, and the system orchestrates them via Temporal Schedules.

Engine note: Scheduling is production-ready for Temporal only. Local LangGraph / ForegroundRunner gets a NoOpScheduleEngine that rejects create/trigger with a clear error so non-Temporal modes do not silently create dead schedules. See multi-agent/lib/mas/scheduling/README.md.

**Backend**

- Introduce mas.scheduling domain module: WorkflowSchedule (composition with Prompt), ScheduleEngine port, repository, and WorkflowScheduleService (CRUD + pause/resume/trigger + finite-run completion + engine reconciliation)
- Add Flask /api/schedules endpoints (schedule.create|update|list|get|pause|resume|trigger|delete|runs)
- Extract ScheduledRunOps + ScheduledSessionRunner (same protocol/runner pattern as BackgroundSessionRunner) so lifecycle ordering lives in the domain; Temporal adapter stays thin
- Temporal inbound: ScheduledSessionWorkflow + consolidated activities (provision_scheduled_session, record_scheduled_outcome)
- Temporal outbound: TemporalScheduleAdapter implementing ScheduleEngine (create/pause/resume/delete/update/trigger/describe)
- Mongo: MongoWorkflowScheduleRepository on workflow_schedules (fields: nested prompt, engine_handle)
- Inject NoOpScheduleEngine when engine_name != temporal; extend session service for schedule-spawned runs
- Register components in the DI container and Temporal worker

**Frontend**

- Add `SchedulePromptModal` — a multi-step modal for creating and editing schedules with recurrence options (cron, interval, custom)
image

image

- Add `ScheduledWorkflows` page — a dedicated view listing all scheduled prompts with status, next-run time, run history, and inline sparkline charts
image
eye icon - opens full prompt and link to session run(s)
image

- Add schedule badge and quick-action dropdown to the `WorkflowsPanel` so users can see and manage schedules without leaving the workflows view
image

- Add sidebar navigation entry and route wiring for the new page
- Add `use-scheduled-blueprints` hook and `prompts` API module for data fetching

**Misc**

- Add a one-time migration script to enforce unique workflow names per blueprint


multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    @workflow.run
+    async def run(self, params: ScheduledSessionParams) -> str:
+        run_id = await workflow.execute_activity(
+            "create_scheduled_session",
+            params,
+            start_to_close_timeout=_ACTIVITY_TIMEOUT,
+            retry_policy=_ACTIVITY_RETRY,
+        )
```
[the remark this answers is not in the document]
shanitzvii: mark_completed non-idempotent - Already guards with schedule_status != ACTIVE check — subsequent calls are no-ops
 _ACTIVITY_RETRY policy - Retry policies and idempotency are complementary — retries handle transient failures, idempotency makes them safe. Added comments noting the guarantee instead.

multi-agent/lib/mas/prompts/service.py
```diff
@@ -0,0 +1,309 @@
...
+
+        prompt = ScheduledPrompt(
+            blueprint_id=blueprint_id,
+            identity=identity,
+            text=text,
+            inputs=inputs or {},
+            source=prompt_source,
+            schedule=schedule_def,
+        )
+        self._repo.save(prompt)
```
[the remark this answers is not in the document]
shanitzvii: the race is real but the fix requires an architectural change (transactional save or DB-level constraint) that is disproportionate to the risk of a soft-limit TOCTOU on a narrow window.

multi-agent/lib/mas/session/service.py
```diff
@@ -237,6 +237,50 @@ def get_chat(self, run_id: str) -> SessionChat:
...
+        return [
+            {
+                "session_id": d.get("run_id"),
+                "status": d.get("status", "UNKNOWN"),
+                "started_at": d.get("run_context", {}).get("started_at"),
+                "metadata": d.get("metadata", {}),
+            }
+            for d in docs
+        ]
+
```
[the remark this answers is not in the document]
shanitzvii: Both get_runs_by_schedule and get_session_detail are serialization-boundary methods — they assemble dicts that are immediately passed to Flask's jsonify() and sent over the wire. Their callers never inspect typed fields programmatically; they just serialize the dict to JSON.

Introducing Pydantic response models here would:

Add a deserialization→re-serialization round-trip (manager returns raw docs/records → construct Pydantic model → .model_dump() → jsonify) with no consumer benefiting from the typed intermediate.
Create coupling between the session domain's internal response models and the prompt service, which calls get_runs_by_schedule and would need to import the new type.
Not fix any bug or prevent any real failure — the current dicts are correct and match the API contract.

ui/client/src/hooks/use-scheduled-blueprints.ts
```diff
@@ -0,0 +1,50 @@
+import { useQuery } from "@tanstack/react-query";
+import { listScheduledPrompts } from "@/api/prompts";
+
+/**
+ * Returns the set of blueprint IDs that have at least one active scheduled prompt
+ * for the given identity. Shares the "scheduled-prompts" query cache with
+ * ScheduledWorkflows so cross-component invalidation works automatically.
+ */
+export function useScheduledBlueprints(
```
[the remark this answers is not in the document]
shanitzvii: The hooks in use-scheduled-blueprints.ts accept userId and identityType as parameters, Both callers already derive these values from useWorkspaceIdentity(). So in practice there's no inconsistency today - all callers already use useWorkspaceIdentity() and pass its values down.

multi-agent/lib/mas/session/service.py
```diff
@@ -42,15 +42,21 @@ def create(
         identity: Identity,
         blueprint_id: str,
         metadata: Dict[str, Any] | SessionMeta | None = None,
+        *,
+        run_id: str | None = None,
     ) -> str:
```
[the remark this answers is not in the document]
shanitzvii: Skipping this fix — not a valid issue.

Both files use str | None (PEP 604 union syntax) for the run_id parameter:


service.py
Lines 43-50
def create(
    self,
    identity: Identity,
    blueprint_id: str,
    metadata: Dict[str, Any] | SessionMeta | None = None,
    *,
    run_id: str | None = None,
) -> str:

user_session_manager.py
Lines 67-74
def create_session(
        self,
        identity: Identity,
        blueprint_id: str,
        metadata: SessionMeta = None,
        *,
        run_id: str | None = None,
) -> str:
Reasons for skipping:

str | None is valid at runtime for this project. The pyproject.toml declares requires-python = ">=3.11", and PEP 604 union syntax has been valid since Python 3.10.
It's the established codebase convention. A search across multi-agent/lib/mas shows str | None / int | None used in 14+ files — this is the project's preferred style, not an oversight.
Optional is already imported in both files but is used for other parameters (e.g., constructor kwargs). Mixing Optional[str] and str | None in the same file wouldn't improve consistency — it would reduce it.

multi-agent/adapters/outbound/temporal/schedule_adapter.py
```diff
@@ -0,0 +1,245 @@
...
+        client = await get_temporal_client()
+        handle = client.get_schedule_handle(temporal_schedule_id)
+        desc = await handle.describe()
+        state = desc.schedule.state
+        exhausted = state.limited_actions and state.remaining_actions == 0
+        return ScheduleInfo(
+            paused=state.paused,
+            remaining_actions=state.remaining_actions if state.limited_actions else None,
+            running=not exhausted,
+        )
```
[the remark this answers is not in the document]
shanitzvii: The _describe method already detects exhaustion for end_at-only schedules via the not desc.info.next_action_times check. The ScheduleInfo.running field already serves as the "end of schedule" indicator, and both _reconcile_batch() and _reconcile_if_needed() already consume it correctly. No change is needed.

multi-agent/lib/mas/scheduling/models.py
```diff
@@ -0,0 +1,100 @@
...
+    blueprint_id: str
+    identity: Identity
+    inputs: Dict[str, Any] = Field(default_factory=dict)
+    source: PromptSource = PromptSource.MANUAL
+    schedule: ScheduleDefinition
+    schedule_status: ScheduleStatus = ScheduleStatus.ACTIVE
+    temporal_schedule_id: Optional[str] = None
+    completed_at: Optional[datetime] = None
+    run_stats: RunStats = Field(default_factory=RunStats)
+    credential_user_id: str = ""
```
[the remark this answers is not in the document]
shanitzvii: Added _RESPONSE_EXCLUDE = {"credential_user_id"} constant and applied exclude=_RESPONSE_EXCLUDE to all model_dump() calls in prompts.py endpoints (6 call sites), plus the list_enriched method in service.py. The field remains on the model and is still persisted to Mongo and passed through Temporal.

multi-agent/adapters/inbound/flask/endpoints/schedules.py
```diff
@@ -0,0 +1,222 @@
...
+    try:
+        svc = current_app.container.prompt_service
+        prompt = svc.create(
+            identity=identity,
+            blueprint_id=blueprint_id,
+            text=text,
+            inputs=inputs,
+            source=source,
+            schedule=schedule,
+            credential_user_id=getattr(g, "identity_username", ""),
```
MayaCrmi, line 45: just interested, why in here specifically do we need this in addition to the identity itself being sent?

multi-agent/adapters/outbound/mongo/blueprint_repository.py
```diff
@@ -39,22 +53,28 @@ def save(self, identity: Identity, spec: BlueprintDraft,
...
+                    "rid_refs": rid_refs,
+                    "updated_at": datetime.now(timezone.utc),
+                }}
+            )
+        except DuplicateKeyError:
+            raise BlueprintDuplicateNameError(spec.name) from None
         return res.modified_count == 1
     
-    def set_metadata(self, *, blueprint_id: str, metadata: Dict[str, Any]) -> bool:
+    def set_metadata(self, *, blueprint_id: str, metadata: dict[str, Any]) -> bool:
```
MayaCrmi, line 77: what caused the change from the Dict and List from typing to a new way of typing here?

multi-agent/adapters/outbound/mongo/scheduled_prompt_repository.py
```diff
@@ -0,0 +1,152 @@
...
+        if not doc:
+            raise KeyError(f"No scheduled prompt with id={prompt_id}")
+        return self._deserialize(doc)
+
+    def update(self, prompt: ScheduledPrompt) -> bool:
+        now = datetime.now(timezone.utc)
+        doc = prompt.model_dump(mode="json")
+        doc["updated_at"] = now
+        doc.pop("created_at", None)
+        if prompt.completed_at is not None:
```
MayaCrmi: the 'completed_at' is set both here and under save so not sure what exactly it means? how come it can be updated in both places? are they both called when a run is finished?

multi-agent/adapters/temporal/models.py
```diff
@@ -89,3 +92,39 @@ class FailSessionParams(BaseModel):
...
+
+class ScheduledSessionParams(BaseModel):
+    """Input to ScheduledSessionWorkflow (triggered by Temporal Schedule)."""
+    prompt_id: str
+    blueprint_id: str
+    identity: Identity
+    text: str = ""
+    inputs: Dict[str, Any] = Field(default_factory=dict)
+    source: PromptSource = PromptSource.MANUAL
+    idempotency_key: Optional[str] = None
```
MayaCrmi: it's a bit קטנוני but do you think there's maybe an 'eaiser' word than  idempotency for this action? i noticed it in the other file and it just feels like a really big word to use here haha so wondering if you think there's something more intuitive for us?

multi-agent/lib/mas/prompts/service.py
```diff
@@ -0,0 +1,425 @@
...
+        credential_user_id: str = "",
+    ) -> ScheduledPrompt:
+        try:
+            bp_doc = self._blueprint_service.get_blueprint_draft_doc(blueprint_id)
+        except KeyError:
+            raise BlueprintNotFoundError(blueprint_id)
+        if not bp_doc.identity.owns(identity):
+            raise PromptPermissionError(blueprint_id, identity)
+
+        active_count = self._repo.count_active_by_blueprint(blueprint_id)
```
MayaCrmi: this is where we check the counter, where do we increase it? like if this is where we create and save it

multi-agent/lib/mas/session/domain/dto.py
```diff
@@ -23,5 +24,6 @@ def from_doc(cls, doc: Mapping[str, Any], blueprint_exists: bool = True, public_
             started_at=rc.get("started_at") or "",
             last_active_at=rc.get("last_active_at") or "",
             blueprint_id=doc.get("blueprint_id", ""),
-            blueprint_exists=blueprint_exists
+            blueprint_exists=blueprint_exists,
+            status=doc.get("status", ""),
```
MayaCrmi, line 28: the new status here is for running session in general, not specifically scheduled ones?

multi-agent/lib/mas/session/service.py
```diff
@@ -42,15 +45,21 @@ def create(
...
         """
         Create a new session record and return its run_id.
         Lightweight — no graph compilation or blueprint resolution.
+
+        When *run_id* is supplied the session is created with that
```
MayaCrmi, line 60: it says here 'when run_id' is supplied but is it optional? when will we not get this?

multi-agent/lib/mas/sharing/cloner.py
```diff
@@ -152,6 +152,7 @@ def clone_blueprint(self, *, blueprint_id: str, ctx: CloneContext) -> Tuple[str,
                 identity=recipient,
                 draft_dict=new_draft.model_dump(mode="json"),
                 metadata=bp_metadata or None,
+                skip_name_check=True,
```
MayaCrmi, line 155: why do we skip the name check here? if i send someone a workflow and they receive it with Name (from mcarmi) and then I send it again, they will have a duplicated name in their repo, no? (both will be workflow_name (from mcarmi))

MayaCrmi: same for sharing with a team - what if we both send the same name workflow to our mutual team?

ui/client/src/components/agentic-ai/RunHistoryPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+  COMPLETED: "bg-emerald-500/15 text-emerald-400 border-emerald-500/30",
+  FAILED: "bg-red-500/15 text-red-400 border-red-500/30",
+  RUNNING: "bg-blue-500/15 text-blue-400 border-blue-500/30",
+};
+
+function badgeClass(status: string): string {
+  return STATUS_BADGE[status] ?? "bg-gray-500/15 text-gray-400 border-gray-500/30";
+}
+
+function formatTime(iso: string): string {
```
MayaCrmi: we have other places where we format time, please make sure this code isn't a duplicate/find a way to consolidate the others with this to be used from one single source

ui/client/src/components/agentic-ai/RunSparkline.tsx
```diff
@@ -0,0 +1,86 @@
...
+  COMPLETED: "bg-emerald-400",
+  FAILED: "bg-red-400",
+  RUNNING: "bg-blue-400 animate-pulse",
+};
+
+function dotColor(status: string): string {
+  return STATUS_COLORS[status] ?? "bg-gray-500";
+}
+
+function formatRelativeTime(iso: string): string {
```
MayaCrmi: same comment as with formatTime

ui/client/src/components/agentic-ai/SchedulePromptModal.tsx
```diff
@@ -0,0 +1,933 @@
...
+  isOpen: boolean;
+  onClose: (saved?: boolean) => void;
+  blueprintId: string;
+  blueprintName: string;
+  userId?: string;
+  identityType?: string;
+  editPrompt?: ScheduledPromptResponse | null;
+}
+
+const DAY_NAMES = ["S", "M", "T", "W", "T", "F", "S"];
```
MayaCrmi: let's also extract this to an outside constant file so we can use it in other places in the future

ui/client/src/components/agentic-ai/SchedulePromptModal.tsx
```diff
@@ -0,0 +1,933 @@
...
+    case "every_week":
+      return { ...base, cron_expression: `${M} ${H} * * ${dayOfWeek}` };
+    case "every_month":
+      return { ...base, cron_expression: `${M} ${H} ${dayOfMonth} * *` };
+    default:
+      return { ...base, interval: "PT1H", remaining_actions: 1 };
+  }
+}
+
+function parseISODuration(iso: string): { value: number; unit: "minute" | "hour" | "day" } | null {
```
MayaCrmi: if when checking what's happening with the other time formatting functions we'll have a time formatting file, move this to be part of said file as well :)

ui/client/src/components/agentic-ai/SchedulePromptModal.tsx
```diff
@@ -0,0 +1,933 @@
...
+          </Button>
+          <Button onClick={handleDone}>Done</Button>
+        </DialogFooter>
+      </DialogContent>
+    </Dialog>
+  );
+}
+
+// ────────────────────────────────────────────────────────────────────────────────
+// Main Component
```
MayaCrmi: it's a very big component, check if maybe there's a way to make it nicer and easier to understand like extracting things to a hook/helper files. if not that's fine, just think it can make this more sustainable

ui/client/src/components/layout/Sidebar.tsx
```diff
@@ -285,7 +285,15 @@ export default function Sidebar() {
             icon={<FaComment className="sidebar-icon" />} 
             label={viewMode === 'team' ? 'Collaboration Hub' : 'Agentic Chats'}
             to="/agentic-chats"
-            isActive={location === '/agentic-chats'}
+            isActive={location === '/agentic-chats' || location.startsWith('/agentic-chats/')}
+            status={null}
+            isCollapsed={isCollapsed}
+          />
+          <NavItem
```
MayaCrmi, line 292: not 100% sure it should be the last in the sidebar, maybe it should be under the other workflows tab and the chats should be at the bottom? what do you think?

ui/client/src/pages/ScheduledWorkflows.tsx
```diff
@@ -0,0 +1,642 @@
...
+import { useScheduledBlueprintCounts } from "@/hooks/use-scheduled-blueprints";
+import { parseUtcDate, cronTimeToLocal } from "@/utils/dateUtils";
+
+const MAX_ACTIVE_SCHEDULES_PER_WORKFLOW = 10;
+
+// ---------------------------------------------------------------------------
+// Schedule label derivation – times are converted to the user's local timezone
+// ---------------------------------------------------------------------------
+
+const DAY_CRON_NAMES = ["SUN", "MON", "TUE", "WED", "THU", "FRI", "SAT"];
```
MayaCrmi: note when you move what i requested this could be called from there :)

ui/client/src/pages/ScheduledWorkflows.tsx
```diff
@@ -0,0 +1,642 @@
...
+
+// ---------------------------------------------------------------------------
+// Status badge
+// ---------------------------------------------------------------------------
+
+type ScheduleStatusType = "active" | "paused" | "completed";
+
+function statusColor(s: ScheduleStatusType): string {
+  switch (s) {
+    case "active":
```
MayaCrmi: do you think this could be used also from an outside file, to be called also from RunHistoryPanel for example? Or are they different css?

ui/client/src/utils/dateUtils.ts
MayaCrmi, line 1: Perfect to have dateUtils, let's have everything I requested here/in a parallel timeUtils file etc (which I think might already exist?? not sure)

MayaCrmi (commented):
Great job with this! Super important and useful ticket 👸🏻

multi-agent/adapters/inbound/flask/endpoints/schedules.py
shanitzvii: its because of scheduled workflows running in team mode, that use MCP servers that require user-specific OAuth tokens (sign-in method, github for example)
when in manual flow (what we had so far) - the entire preprocessing happens in flask process of mas, but in scheduled - its in temporal worker, so i had to inject, "mock" the parameters within the temporal worker activity.
for those user-specific OAuth MCPs, 
in the flask, endpoint receives identity.id (the individual user's ID). It passes it as logged_in_user to input_projector.apply(). That stamps the last person who spoke as the credential_user_id in the session's run_context. So the OAuth token used is always from whoever last triggered the workflow.

in scheduled workflow, i needed to store the credential user ID from the flask endpoint and propagate it properly to ScheduledSessionParams so it's available at activity execution time.

very specific usecase, can explain more if neeeded :)

multi-agent/adapters/outbound/mongo/blueprint_repository.py
MayaCrmi: ok I think I understand!

multi-agent/adapters/inbound/flask/endpoints/schedules.py
MayaCrmi: I think I get it, though I still don't entirely understand why you need to inject that specifically instead of using the Identity field we're already sending here? 
Like, when it's from the UI the identity is the team, and then we call the logged in user to the username in this. But in the API I don't see the logged in user so I need to send it explicitly?
Did I get it right?

multi-agent/adapters/outbound/mongo/scheduled_prompt_repository.py
shanitzvii: save is called only once -  in initial creation of a scheduled prompt, when creating ,completed_at will be none so the TTL index in mongo (7-day expiry) won't start.

while update is called and used for all further modifications to existing prompt, including the completion of the workflow.

The completed_at guard in save is defensive code, in practice only the update path sets it
i can remove from save, its dead code.

multi-agent/lib/mas/session/domain/dto.py
shanitzvii: yep, it applies to any running session, it reads whatever status value is stored in the session document.
But, i only use this data for the scheduled workflows -  to support the Scheduled Workflows page, and display the status of sessions that were triggered by scheduled prompts

multi-agent/adapters/outbound/mongo/scheduled_prompt_repository.py
MayaCrmi: Yep that makes more sense to me! I feel like we can remove the guard :) If you agree you can resolve this thread once it's removed

multi-agent/lib/mas/session/domain/dto.py
MayaCrmi: So in my sessions page I will see "Active" on sessions that were scheduled, alongside a label that says they were scheduled, but for running sessions that were triggered manually I won't see the "Active" status?

multi-agent/lib/mas/session/service.py
shanitzvii: run_id parameter is only provided by the scheduled-prompt Temporal workflow.
regular sessions let the user_session_manager auto-generate a UUID (when user starts new chat session from ui)
while in schedule wf , the session is created within temporal worker , and we need this paramater as deterministic in case of failures and retries of the create_scheduled_session activity, so we won'thave duplicate session in the database.

multi-agent/adapters/inbound/flask/endpoints/schedules.py
```diff
@@ -0,0 +1,222 @@
...
+        return jsonify({"error": str(e), "error_type": "BLUEPRINT_NOT_FOUND"}), 404
+    except PromptPermissionError as e:
+        return jsonify({"error": str(e), "error_type": "FORBIDDEN"}), 403
+    except PromptLimitExceededError as e:
+        return jsonify({"error": str(e), "error_type": "LIMIT_EXCEEDED"}), 409
+    except ValueError as e:
+        return jsonify({"error": str(e), "error_type": "VALIDATION_ERROR"}), 400
+    except Exception:
+        logger.exception("Unhandled error in create_prompt")
+        return jsonify({"error": "Internal server error", "error_type": "INTERNAL_ERROR"}), 500
```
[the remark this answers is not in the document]
shanitzvii: every single endpoint file uses inline try/except inside each controller function.
every domain repeats the same try: ... except SomeDomainError: ... except Exception: return jsonify({"error": ...}), 500 boilerplate inside each controller function. No delegation whatsoever.

resolving since the pattern aligns with codebase.

multi-agent/lib/mas/prompts/service.py
shanitzvii: im dont hold a counter, i query the DB directly, and count.
for now we forced max 10 active per blueprint, so I thought i dont need to have a counter. also, the prompts collection is saved per scheduled prompt object, which has a field of blueprint_id it attached ->
its the opposite way of counting, thats why i prefered to query directly, filter and count.

multi-agent/lib/mas/sharing/cloner.py
shanitzvii: nope
one of the things i added here is we wont have duplicated names for blueprints in he repo anymore (we discussed that a bit in the adr)

so in shared wf,
he will have workflow_name (from x) (n)
n is the n'th copy of this workflow for the user

multi-agent/lib/mas/prompts/service.py
MayaCrmi: oh oh okay! no problem :)

multi-agent/lib/mas/session/service.py
MayaCrmi: alright, so it's only called for scheduled workflows, right?
is the argument set as Optional? i can't see it in the snippet of code i commented on and just want to make sure. if so you can just resolve this :)

multi-agent/lib/mas/sharing/cloner.py
MayaCrmi: okay if you're sure about it sounds good! like i currently have on my PP two workflows with the same name with the (from X) suffix so just wanted to make sure

ui/client/src/pages/ScheduledWorkflows.tsx
shanitzvii: mmm i think calling it from RunHistoryPanel would be wrong cause its slightly different css and different concepts, statusColor colors schedule's lifecycle state while the badge in RunHistoryPanel colors individual run context status

i can exract the badge markup pattern while keeping the two separate color-mapping functions.
do you think its necessary?

ui/client/src/components/layout/Sidebar.tsx
shanitzvii: mmm i also התלבטתי 
i saw logic in keeping it last, since we can jump to the sessions (agentic chat page) from the scheduled workflows page, so it make sense it will be under.
 wdyt?

MayaCrmi: mmmm i see what you're saying
honestly i still stand by what i said in our meeting and think the naming is a bit confusing?? like it should be like scheduled prompts instead of workflows in the title as well idk... and then it feels like it makes more sense not straight under the workflows
honedtly i know it's not that big of a deal lol so i support your decision haha whatever you think!

ui/client/src/pages/ScheduledWorkflows.tsx
MayaCrmi: my train of thought is i want whoever expands or adds to this feature to be able to use the same css and style as what you added easily. so i'd want all of these to be in a general file where it can be called, insetad of calling the same css style like 500-blue... from different files for different purposes. if there's a way to make it more single-source-of-truth style, i think it would be better, but if there's vastly differnet and it would just be complicated i'm fine with you keeping it that way.

ui/client/src/utils/dateUtils.ts
shanitzvii: added the time utils into the date utils
with all functions mentioned before, with mapping existing functions to use this utils. :)

ui/client/src/components/layout/Sidebar.tsx
shanitzvii: haha i dont know i prefer workflows over prompts but that really no biggy
i think i'd keep it that way :)

ui/client/src/components/agentic-ai/SchedulePromptModal.tsx
shanitzvii: done!

multi-agent/adapters/temporal/models.py
shanitzvii: changed to 
dedupe_key <3

multi-agent/adapters/inbound/flask/endpoints/schedules.py
shanitzvii: We need it here mostly because of team mode.
The core issue is that in team mode, Identity.id is the team ID, not any individual user's ID. OAuth tokens are per-person ,no "team OAuth token".
So the paramater we need-  whose OAuth token we use when the workflow runs on behalf of a team.

In the flask manual flow: 
it's a two-step process - 
1) Session CREATE - created with team identity, all get ownership
2) Session SUBMIT - what actuallyfires the temporal workflow ,real user making an HTTP request in "live" (by clicking run in the UI) . the request carries the userId (from UI - SubmitSessionParams) ,flask validates the session and decorator falls to user, and submit endpoint runs with identity.id of the user as loggedin user, (not team).
then in input_projector.apply() → record.run_context.with_credential_user("user"), which sets the tag of "credential_user_id" in ExecutionContext.


In the scheduled/Temporal flow there is no human present at execution time, only in creation time (API).
there's only one request :prompt.create. which must simultaneously
1)set team ownership 
2)capture the human for credentials
 so we dont use the submit api endpoint. an inner temporal cron timer fires, temporal picks up the job- and by that time there is no HTTP submit request, no auth headers, no browser session. Nobody is "logged in" when the schedule triggers, there's nothing to extract credential_user_id from. If we only sent Identity, we'd only have the team ID, and we'd have no way to know which team member's OAuth token to use.

because of that, we need to capture it at schedule-creation time (when the user is present and creating the schedule via endpoint) and store it on the prompt model so it's available later when Temporal runs.


In personal mode this distinction doesn't matter because identity.id already is the user. It's specifically the team + OAuth + scheduled combination that requires it

multi-agent/adapters/outbound/mongo/scheduled_prompt_repository.py
shanitzvii: done - no guard in save :)

multi-agent/adapters/inbound/flask/endpoints/prompts.py
```diff
@@ -0,0 +1,222 @@
...
+    except PromptNotFoundError as e:
+        return jsonify({"error": str(e), "error_type": "NOT_FOUND"}), 404
+    except PromptPermissionError as e:
+        return jsonify({"error": str(e), "error_type": "FORBIDDEN"}), 403
+    except Exception:
+        logger.exception("Unhandled error in get_prompt")
+        return jsonify({"error": "Internal server error", "error_type": "INTERNAL_ERROR"}), 500
+
+
+@prompts_bp.route("/prompt.schedule.pause", methods=["POST"])
```
Nirsisr: Shani's, this comment relevant for both `pause`, `resume`, `trigger` all the ENDPOINTS which basically contains schedule.

I think we talk about it in the ADR, as I see it, Prompt is a class/representation that can be consumed by schedule or blueprints as an entity, therefore the entire methods connected to the life-cycle of the `schedule-workflow` I would chose putting them:
1) New schedule_bp (endpoints)
2) Under sessions (since we basically resume/pause/..) sessions by using the schedule service.

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
```diff
@@ -0,0 +1,123 @@
...
+                params.run_id,
+                params.started_at,
+                exc_info=True,
+            )
+            started = datetime.now(timezone.utc)
+
+        self._prompt_service.record_run(
+            params.prompt_id, params.run_id, params.status, started,
+        )
+        self._prompt_service.mark_completed(params.prompt_id)
```
Nirsisr: Shani follows my previous comment, I feel that we should record on the _schedule_, we will both record the prompt but as part of the scheduled workflow.

Nirsisr: Same for mark_completed logic of course.

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
Nirsisr, line 1: @shanitzvii In whatever READMEs you are documenting the great Schedule addition, make sure to mention that it exist on TEMPORAL engine only ATM (production orientated). 

I didn't read yet the entire CR but I guess you didn't implement it for ForeGroundRunner with LangGraph since it will require implementing there schedule with Celery/Kafka/etc. So it's something we need to clarify.

If someone want to implement schedule when other engine is on (LangGraph for local work) it just need to me clear what methods expected being implemented (the abstraction layer, e.g. create_scheduled_session and more)

.cursor/skills/pipeline/phases/arch-reviewer.md
```diff
@@ -192,6 +192,7 @@ A finding should only be MAJOR or CRITICAL if it is tagged `[NEW]` — meaning *
 - Do NOT approve if architectural violations or unverified claims exist.
 - If the diff contains only non-Python files (CI configs, docs, markdown), acknowledge that hexagonal rules do not apply and focus on correctness, consistency, and codebase conventions instead.
 - Machine-generated files (`**/pnpm-lock.yaml`, `**/package-lock.json`, `**/yarn.lock`, `**/*.lock`, `**/*.generated.*`) must be skipped entirely — do NOT review or flag them for size, style, or content.
+- **Fix Convention Verification**: Before emitting a `**Fix →**` block, verify the proposed mechanism (e.g., adding logging, introducing a base class, changing a pattern) is already used in sibling files in the same architectural layer. If siblings don't use it, adjust the fix to match existing conventions or explicitly note the deviation.
```
sfiresht, line 195: like ++

multi-agent/adapters/inbound/flask/endpoints/blueprints.py
```diff
@@ -182,6 +183,8 @@ def save_blueprint(identity, blueprint_raw=None, metadata=None):
             "blueprint_id": blueprint_id
         }), 201
 
+    except BlueprintDuplicateNameError as e:
```
sfiresht, line 186: don't we allow the same naming anymore? or did you run into this ?

multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
```diff
@@ -0,0 +1,147 @@
+"""
```
sfiresht, line 1: generic question - couldn't we reuse the sessionworkflow and add optional params where needed? (of course that extra functions are needed but isn't the base the same - workflow.run).
that way you have one code base/mongo collection for all and filter by type or somthing?

sfiresht (commented):
looks nice, I think I missd the ADR so some questions are related to it.

multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
```diff
@@ -0,0 +1,147 @@
...
+    BuildSessionWorkflowParamsInput,
+    PostExecutionParams,
+    RunOutcome,
+    ScheduledSessionParams,
+    SessionWorkflowParams,
+    StageScheduledInputsParams,
+)
+
+with workflow.unsafe.imports_passed_through():
+    from inbound.temporal.workflows.session_workflow import SessionWorkflow
```
Nirsisr: @shanitzvii Can you explain the `with` section here, what does it stand for (with the imports pass)?

multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
Nirsisr, line 1: @shanitzvii You are rockstar, I enjoyed read it, all documented, code separated into logical steps, very organized and clean. Well done!

multi-agent/adapters/outbound/mongo/workflow_schedule_repository.py
```diff
@@ -0,0 +1,150 @@
...
+                ("updated_at", pymongo.DESCENDING),
+            ],
+            background=True,
+        )
+        self._col.create_index("blueprint_id", background=True)
+        self._col.create_index("schedule_status", background=True)
+        try:
+            self._col.create_index(
+                "completed_at",
+                expireAfterSeconds=604800,  # 7-day TTL
```
Nirsisr, line 40: @shanitzvii We are enable scheduling only on up-coming 7 days interval?

multi-agent/adapters/outbound/temporal/schedule_adapter.py
Nirsisr, line 1: @shanitzvii I understand why you did it, but the implementation of both pause/resume/delete within the _pause/_resume/_delete in the same CLASS is super confusing.

`pause` called to  `_pause` (same for the other methods) I got confused here.

multi-agent/bootstrap/container.py
```diff
@@ -396,6 +413,13 @@ def _create_background_engine(engine_name: str):
             return TemporalSessionEngine()
         return None
 
+    @staticmethod
+    def _create_schedule_adapter(engine_name: str):
+        if engine_name == "temporal":
+            from outbound.temporal.schedule_adapter import TemporalScheduleAdapter
+            return TemporalScheduleAdapter()
+        return None
```
Nirsisr: Related to previous comment, we are now supporting schedule only on temporal, we need to make sure it doesn't break if someone ran locally with LangGraph.

I guess no mather where you run you still see in the GUI (Schedule Workflows page) it's not rendered pair the selected ENGINE in the backend, so user can fill stuff and same APIs will be called. 
If the engine is not temporal we will raise ERROR message? How does it work?

I just encounter the same issue with `tracing_service` therefore I implemented the NoOp() which implemented the ABSTRACTION of how tracing service should looks like but all the methods return PASS (since I didn't want it will break other modes) so asking it here to understand how you tackle it

multi-agent/adapters/inbound/flask/endpoints/blueprints.py
shanitzvii: no, in the ADR we discussed that we'll allow only unique names, i implemented it as part of the PR and will run migration script on deployment.

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
shanitzvii: done! thanks for noticing me :) 
it's under lib/mas/scheduling/README.md

multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
shanitzvii: i now ewimplemented it as a wrapper workflow that delegates the run to the session workflow runner :)

multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+
+from temporalio import workflow
+from temporalio.common import RetryPolicy
+from temporalio.exceptions import ChildWorkflowError, is_cancelled_exception
+
+from mas.graph.state.graph_state import GraphState
+from mas.scheduling.models import RunOutcome
+from mas.session.domain.exceptions import ScheduledSessionCancelledException
+from mas.session.execution.ports import ScheduledExecutionParams
+from mas.session.execution.scheduled_runner import ScheduledSessionRunner
```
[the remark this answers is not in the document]
shanitzvii: its the same convention as session_workflow.py, it's ok

multi-agent/adapters/outbound/mongo/workflow_schedule_repository.py
shanitzvii: no, there is no time limit. the TTL is for what we keep under "scheduled_workflows" in mongo after they fully completed their runs. it will be shown for 7 days as "completed", and after that we'll delete it from mongo (and from the scheduled workflows page as outcome)

multi-agent/lib/mas/scheduling/service.py
```diff
@@ -0,0 +1,492 @@
...
+            raise
+        wf_schedule = wf_schedule.model_copy(update={"engine_handle": handle})
+        if not self._repo.update(wf_schedule):
+            try:
+                self._schedule_engine.delete(handle)
+            except EngineScheduleNotFoundError:
+                logger.info(
+                    "Engine schedule %s already absent during cleanup for schedule %s; no deletion needed",
+                    handle, wf_schedule.id,
+                )
```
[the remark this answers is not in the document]
shanitzvii: the codebase does not currently have a “services never log; use decorators” pattern. Sibling services (session, blueprints, auth, …) all call logger directly, and schedule Flask endpoints already log unhandled exceptions at the adapter edge.

odaiodeh:
@shanitzvii 
# Scheduled Workflow Execution — Architecture Redesign

## Status: Proposed

---

## 1. Problem Statement

The current `ScheduledSessionWorkflow` implementation bakes domain logic (session
creation, input staging, blueprint resolution, run recording) directly into Temporal
activities. This violates the project's hexagonal architecture by:

1. **Coupling domain logic to Temporal** — adding a second scheduler backend
   (Celery, APScheduler) requires rewriting all orchestration logic.
2. **Breaking the established pattern** — normal session execution uses
   `BackgroundSessionRunner` + `BackgroundSessionOps` protocol.  The scheduled
   path bypasses this entirely.
3. **Naming inconsistency** — backend calls it `ScheduledPrompt`, UI calls it
   "Scheduled Workflows".  The entity represents a recurring workflow execution
   configuration, not a scheduled piece of text.

---

## 2. Naming Change

### Current → Proposed

| Layer | Current | Proposed |
|-------|---------|----------|
| Domain model | `ScheduledPrompt` | `WorkflowSchedule` |
| Domain service | `PromptService` | `WorkflowScheduleService` |
| Domain module | `lib/mas/prompts/` | `lib/mas/scheduling/` |
| Repository | `ScheduledPromptRepository` | `WorkflowScheduleRepository` |
| Mongo collection | `scheduled_prompts` | `workflow_schedules` |
| Flask blueprint | `prompts_bp` | `schedules_bp` |
| API endpoints | `/prompt.*` | `/schedule.*` |
| Port (timer mgmt) | `SchedulePort` | `ScheduleEngine` |
| UI page route | `/scheduled-workflows` | `/scheduled-workflows` (unchanged) |

### Rationale

- **`WorkflowSchedule`**: the entity is a configuration that causes a workflow
  (blueprint) to execute on a recurring schedule.  The prompt text is one field
  among many (inputs, overlap policy, timezone, etc.).
- Aligns backend naming with the UI's existing `/scheduled-workflows` route.
- "Schedule" is the noun (the configuration); "Workflow" is what's being scheduled.
- `Session` is the *output* — each tick produces a session.  You don't "schedule
  a session"; you schedule something that *creates* sessions.

---

## 3. Architecture: Protocol Composition (Decorator Pattern)

### 3.1 Existing Pattern (unchanged)

```
BackgroundSessionRunner (domain — canonical ordering)
    │
    │   async run(ops: BackgroundSessionOps)
    │       ops.begin()              → mark RUNNING
    │       ops.execute_graph(state) → run the graph
    │       ops.complete(state)      → mark COMPLETED
    │       (or ops.fail / ops.cancel)
    │
    └── SessionWorkflow implements BackgroundSessionOps (Temporal adapter)
```

This handles **one session execution**.  It stays exactly as is.

### 3.2 New Pattern (wraps the existing)

```
ScheduledSessionRunner (domain — canonical ordering for scheduled execution)
    │
    │   async run(ops: ScheduledRunOps)
    │       session_id, params = ops.provision()    → prepare session
    │       outcome = ops.execute(params)           → run via existing machinery
    │       ops.record(session_id, outcome)         → record stats
    │
    │   ┌────────────────────────────────────────────────────────────┐
    │   │ ops.execute() internally delegates to:                      │
    │   │   Temporal → child SessionWorkflow (uses BackgroundRunner)  │
    │   │   Celery   → BackgroundSessionRunner.run(CeleryOps)         │
    │   └────────────────────────────────────────────────────────────┘
    │
    └── ScheduledSessionWorkflow implements ScheduledRunOps (Temporal adapter)
```

**Key principle**: `ScheduledSessionRunner` wraps `BackgroundSessionRunner`.
It adds scheduling-specific pre/post steps without duplicating session lifecycle.

---

## 4. Domain Layer Design

### 4.1 Protocol: `ScheduledRunOps`

```python
# lib/mas/session/execution/scheduled_runner.py

from typing import Protocol, runtime_checkable
from mas.prompts.models import RunOutcome


@runtime_checkable
class ScheduledRunOps(Protocol):
    """
    Engine-specific mechanics for scheduled session execution.

    This protocol does NOT redefine session execution — it wraps it.
    The execute() method internally delegates to the existing
    BackgroundSessionRunner + BackgroundSessionOps machinery.

    Temporal: execute() = child SessionWorkflow
    Celery:  execute() = direct BackgroundSessionRunner.run(CeleryOps)
    """

    async def provision(self) -> tuple[str, "SessionWorkflowParams"]:
        """
        Prepare a session for execution.

        Calls domain services to:
          1. Create session record (PENDING)
          2. Stage inputs + prompt text
          3. Resolve blueprint, compile graph

        Returns (session_id, params_ready_for_execution).
        """
        ...

    async def execute(self, params: "SessionWorkflowParams") -> RunOutcome:
        """
        Run the full session lifecycle via existing machinery.

        Internally uses BackgroundSessionRunner — the adapter decides
        HOW to invoke it (child workflow, direct call, task submission).
        """
        ...

    async def record(self, session_id: str, outcome: RunOutcome) -> None:
        """
        Post-execution: record run stats, conditionally complete
        finite schedules.

        Calls domain services (WorkflowScheduleService.record_run,
        mark_completed).
        """
        ...
```

### 4.2 Runner: `ScheduledSessionRunner`

```python
# lib/mas/session/execution/scheduled_runner.py

class ScheduledSessionRunner:
    """
    Domain-layer orchestrator for scheduled execution.

    Wraps session execution with scheduling pre/post steps.
    The actual session lifecycle is delegated to existing machinery
    via ops.execute() — no duplication.

    Mirrors BackgroundSessionRunner for the scheduled path.
    """

    async def run(self, ops: ScheduledRunOps) -> str:
        """
        Execute the full scheduled session lifecycle.

        Returns session_id of the created run.
        """
        session_id, params = await ops.provision()
        outcome = RunOutcome.FAILED
        try:
            outcome = await ops.execute(params)
        finally:
            await ops.record(session_id, outcome)
        return session_id
```

### 4.3 Domain Service Method: `prepare_for_execution`

The `provision()` step in each adapter should delegate to **one domain service
call** — not reimplement create/stage/resolve logic:

```python
# lib/mas/session/service.py  (addition to existing SessionService)

class SessionService:

    def prepare_for_scheduled_execution(
        self,
        *,
        identity: Identity,
        blueprint_id: str,
        inputs: Dict[str, Any],
        text: str = "",
        metadata: SessionMeta | None = None,
        logged_in_user: str = "",
        run_id: str | None = None,
    ) -> tuple[str, SessionWorkflowParams]:
        """
        Domain use-case: prepare a session for scheduled background execution.

        Creates session → stages inputs → resolves blueprint → returns
        everything needed for execution.

        Used by:
          - ScheduledRunOps.provision() in any engine adapter

        Parallels the existing submit() flow but stops before engine.submit(),
        returning the prepared params instead.
        """
        session_id = self.create(
            identity=identity,
            blueprint_id=blueprint_id,
            metadata=metadata,
            run_id=run_id,
        )
        record = self._manager.get_record(session_id)
        combined_inputs = {**inputs}
        if text:
            combined_inputs["user_prompt"] = text
        self._projector.apply(record, combined_inputs, logged_in_user=logged_in_user)

        session = self._manager.get_session(session_id)
        exec_context = ExecutionContext(
            session_id=session_id,
            identity=identity,
            scope="public",
            engine_name=self._engine_name,
            tags=session.record.run_context.tags,
        )
        graph_params = GraphExecutionParams(
            state=session.graph_state,
            graph_definition=session.executable_graph.graph_definition,
            session_id=session_id,
            execution_context=exec_context,
        )
        params = SessionWorkflowParams(
            run_id=session_id,
            execution_context=exec_context,
            graph_execution_params=graph_params,
        )
        return session_id, params
```

---

## 5. Adapter Implementations

### 5.1 Temporal Adapter (thin wrapper)

```python
# adapters/inbound/temporal/workflows/scheduled_session_workflow.py

@workflow.defn
class ScheduledSessionWorkflow:
    """
    Temporal adapter — implements ScheduledRunOps.

    Each method is a thin activity/child-workflow call.
    Domain logic lives in domain services, not here.
    """

    @workflow.run
    async def run(self, params: ScheduledSessionParams) -> str:
        self._params = params
        runner = ScheduledSessionRunner()
        return await runner.run(self)

    # ── ScheduledRunOps implementation ───────────────────────────

    async def provision(self) -> tuple[str, SessionWorkflowParams]:
        # ONE activity → calls session_service.prepare_for_scheduled_execution()
        return await workflow.execute_activity(
            "provision_scheduled_session",
            self._params,
            start_to_close_timeout=timedelta(minutes=2),
            retry_policy=RetryPolicy(maximum_attempts=3),
            result_type=ProvisionResult,
        )

    async def execute(self, params: SessionWorkflowParams) -> RunOutcome:
        # Delegates to existing SessionWorkflow (which uses BackgroundSessionRunner)
        try:
            await workflow.execute_child_workflow(
                SessionWorkflow.run,
                params,
                id=f"sched-session-{params.run_id}",
                execution_timeout=timedelta(hours=4),
                result_type=GraphState,
            )
            return RunOutcome.COMPLETED
        except ChildWorkflowError:
            return RunOutcome.FAILED

    async def record(self, session_id: str, outcome: RunOutcome) -> None:
        # ONE activity → calls schedule_service.record_run() + mark_completed()
        await workflow.execute_activity(
            "record_scheduled_outcome",
            RecordOutcomeParams(
                prompt_id=self._params.prompt_id,
                session_id=session_id,
                outcome=outcome,
                started_at=workflow.now().isoformat(),
            ),
            start_to_close_timeout=timedelta(seconds=30),
            retry_policy=RetryPolicy(maximum_attempts=3),
        )
```

### 5.2 Celery Adapter (hypothetical, zero domain duplication)

```python
# adapters/inbound/celery/scheduled_task.py

class CeleryScheduledRunOps:
    """Celery adapter — implements ScheduledRunOps with direct domain calls."""

    def __init__(self, params, session_service, schedule_service):
        self._params = params
        self._session_service = session_service
        self._schedule_service = schedule_service

    async def provision(self) -> tuple[str, SessionWorkflowParams]:
        # Same domain call — no reimplementation
        return self._session_service.prepare_for_scheduled_execution(
            identity=self._params.identity,
            blueprint_id=self._params.blueprint_id,
            inputs=self._params.inputs,
            text=self._params.text,
            logged_in_user=self._params.credential_user_id,
        )

    async def execute(self, params: SessionWorkflowParams) -> RunOutcome:
        # Directly uses BackgroundSessionRunner — already inside a task
        ops = CelerySessionOps(params, self._session_service)
        runner = BackgroundSessionRunner()
        try:
            await runner.run(ops)
            return RunOutcome.COMPLETED
        except Exception:
            return RunOutcome.FAILED

    async def record(self, session_id: str, outcome: RunOutcome) -> None:
        # Same domain call
        self._schedule_service.record_run(...)
        self._schedule_service.mark_completed(...)
```

---

## 6. Port Separation

Two distinct ports — different responsibilities, same infrastructure:

### `BackgroundSessionEngine` (existing, unchanged)

```python
class BackgroundSessionEngine(ABC):
    """How do I run ONE session in the background?"""

    def generate_handle(self, session_id: str) -> str: ...
    def submit(self, session: WorkflowSession, request: SubmitSessionRequest) -> None: ...
    def cancel(self, handle: str) -> None: ...
```

### `ScheduleEngine` (renamed from `SchedulePort`)

```python
class ScheduleEngine(ABC):
    """How do I manage WHEN workflow sessions fire?"""

    def create_schedule(self, schedule: WorkflowSchedule) -> str: ...
    def pause(self, temporal_schedule_id: str) -> None: ...
    def resume(self, temporal_schedule_id: str) -> None: ...
    def delete(self, temporal_schedule_id: str) -> None: ...
    def update_schedule(self, temporal_schedule_id: str, schedule: WorkflowSchedule) -> None: ...
    def trigger_now(self, temporal_schedule_id: str) -> None: ...
    def describe(self, temporal_schedule_id: str) -> ScheduleInfo: ...
    def describe_batch(self, schedule_ids: list[str]) -> BatchDescribeResult: ...
```

### Container Wiring

```python
# bootstrap/container.py

if engine_name == "temporal":
    self.session_engine = TemporalSessionEngine()       # BackgroundSessionEngine
    self.schedule_engine = TemporalScheduleAdapter()    # ScheduleEngine
elif engine_name == "celery":
    self.session_engine = CelerySessionEngine()
    self.schedule_engine = CeleryBeatScheduleAdapter()
else:
    self.session_engine = None
    self.schedule_engine = None
```

---

## 7. Composition Diagram

```
┌─────────────────────────────────────────────────────────────────────┐
│  DOMAIN LAYER                                                        │
│                                                                      │
│  ┌─────────────────────┐         ┌──────────────────────────┐      │
│  │ SessionService       │         │ WorkflowScheduleService   │      │
│  │                      │         │                           │      │
│  │ prepare_for_         │         │ create() / update()       │      │
│  │ scheduled_execution()│         │ pause() / resume()        │      │
│  │                      │         │ record_run()              │      │
│  │ (create+stage+resolve│         │ mark_completed()          │      │
│  │  in ONE call)        │         │                           │      │
│  └──────────┬───────────┘         └─────────────┬────────────┘      │
│             │                                    │                    │
│  ┌──────────▼────────────────────────────────────▼──────────────┐   │
│  │  ScheduledSessionRunner                                       │   │
│  │                                                               │   │
│  │    async run(ops: ScheduledRunOps):                            │   │
│  │        session_id, params = ops.provision()   # domain call   │   │
│  │        outcome = ops.execute(params)          # wraps inner   │   │
│  │        ops.record(session_id, outcome)        # domain call   │   │
│  │                                                               │   │
│  │         ┌─────────────────────────────────────────────┐      │   │
│  │         │  BackgroundSessionRunner (EXISTING)          │      │   │
│  │         │    begin → execute_graph → complete/fail     │      │   │
│  │         │    (used inside ops.execute())               │      │   │
│  │         └─────────────────────────────────────────────┘      │   │
│  └───────────────────────────────────────────────────────────────┘   │
│                                                                      │
│  Protocols:                                                          │
│    ScheduledRunOps    → provision / execute / record                  │
│    BackgroundSessionOps → begin / execute_graph / complete / fail     │
│                                                                      │
│  Ports:                                                              │
│    BackgroundSessionEngine → submit / cancel (one-shot execution)     │
│    ScheduleEngine          → create / pause / resume (timer mgmt)    │
└──────────────────────────────────────────────────────────────────────┘
                    ▲                              ▲
                    │ implements                   │ implements
      ┌─────────────┴───────────┐    ┌────────────┴────────────────┐
      │  Temporal Adapters       │    │  (future) Celery Adapters    │
      │                          │    │                              │
      │  ScheduledSessionWorkflow│    │  CeleryScheduledRunOps      │
      │  (ScheduledRunOps)       │    │  (ScheduledRunOps)          │
      │                          │    │                              │
      │  SessionWorkflow         │    │  CelerySessionOps           │
      │  (BackgroundSessionOps)  │    │  (BackgroundSessionOps)     │
      │                          │    │                              │
      │  TemporalScheduleAdapter │    │  CeleryBeatScheduleAdapter  │
      │  (ScheduleEngine)        │    │  (ScheduleEngine)           │
      │                          │    │                              │
      │  TemporalSessionEngine   │    │  CelerySessionEngine        │
      │  (BackgroundSessionEngine)│    │  (BackgroundSessionEngine)  │
      └──────────────────────────┘    └────────────────────────────┘
```

---

## 8. What Lives Where (File Layout)

```
lib/mas/
├── scheduling/                          ← NEW module (renamed from prompts/)
│   ├── __init__.py
│   ├── models.py                        ← WorkflowSchedule, ScheduleDefinition, RunStats
│   ├── repository.py                    ← WorkflowScheduleRepository (ABC)
│   ├── ports.py                         ← ScheduleEngine (ABC), ScheduleInfo
│   └── service.py                       ← WorkflowScheduleService (CRUD + lifecycle)
│
├── session/execution/
│   ├── background_runner.py             ← EXISTING: BackgroundSessionOps + BackgroundSessionRunner
│   ├── scheduled_runner.py              ← NEW: ScheduledRunOps + ScheduledSessionRunner
│   ├── foreground_runner.py             ← EXISTING (unchanged)
│   ├── ports.py                         ← EXISTING: BackgroundSessionEngine
│   ├── lifecycle.py                     ← EXISTING (unchanged)
│   └── input_projector.py              ← EXISTING (unchanged)
│
├── session/service.py                   ← EXISTING + prepare_for_scheduled_execution()
│
└── core/prompt.py                       ← EXISTING: BasePrompt (shared base)

adapters/
├── outbound/temporal/
│   ├── session_engine.py                ← EXISTING: TemporalSessionEngine
│   └── schedule_adapter.py             ← EXISTING: TemporalScheduleAdapter (ScheduleEngine)
│
├── outbound/mongo/
│   ├── session_repository.py            ← EXISTING
│   └── workflow_schedule_repository.py  ← RENAMED from scheduled_prompt_repository.py
│
├── inbound/temporal/
│   ├── workflows/
│   │   ├── session_workflow.py          ← EXISTING (BackgroundSessionOps)
│   │   ├── scheduled_session_workflow.py ← REFACTORED (ScheduledRunOps, thin)
│   │   └── graph_traversal_workflow.py  ← EXISTING
│   └── activities/
│       ├── schedule_activities.py       ← SIMPLIFIED: 2 activities not 4
│       └── ...
│
└── inbound/flask/endpoints/
    └── schedules.py                     ← RENAMED from prompts.py
```

---

## 9. Migration Steps

1. **Rename domain**: `prompts/` → `scheduling/`, `ScheduledPrompt` → `WorkflowSchedule`
2. **Rename API endpoints**: `/prompt.*` → `/schedule.*`
3. **Add `prepare_for_scheduled_execution()`** to `SessionService`
4. **Create `ScheduledRunOps` protocol** + `ScheduledSessionRunner` in domain
5. **Refactor `ScheduledSessionWorkflow`** to implement `ScheduledRunOps`
   - Collapse 3 setup activities into 1 (`provision_scheduled_session`)
   - Collapse post-execution into 1 (`record_scheduled_outcome`)
6. **Rename `SchedulePort`** → `ScheduleEngine` (align with `BackgroundSessionEngine`)
7. **Rename Mongo collection** (migration script: `scheduled_prompts` → `workflow_schedules`)
8. **Update UI API client** (`/prompt.*` → `/schedule.*`)

---

## 10. Key Principles Applied

| Principle | How |
|-----------|-----|
| **Hexagonal architecture** | Domain defines protocols; adapters implement them |
| **Decorator pattern** | `ScheduledSessionRunner` wraps `BackgroundSessionRunner` |
| **Domain owns ordering** | Runners define sequence; adapters supply mechanics |
| **No domain logic in adapters** | Activities call domain services, not reimplement them |
| **Engine-agnostic domain** | Swap Temporal ↔ Celery by implementing protocols |
| **Single domain entry point** | `prepare_for_scheduled_execution()` — one call, reusable |
| **Interface segregation** | `BackgroundSessionEngine` ≠ `ScheduleEngine` — different responsibilities |

---

## 11. Before/After Comparison

### Before (current PR)

```
Temporal fires tick
    → activity: create session (domain logic in adapter)
    → activity: stage inputs (domain logic in adapter)
    → activity: resolve blueprint (domain logic in adapter)
    → child: SessionWorkflow (ok, reuses existing)
    → activity: record run (domain logic in adapter)

5 steps, 4 activities, domain logic scattered in Temporal
```

### After (proposed)

```
Temporal fires tick
    → ScheduledSessionRunner.run(self)
        → activity: provision (1 call to domain service)
        → child: SessionWorkflow (reuses existing BackgroundSessionRunner)
        → activity: record (1 call to domain service)

3 steps, 2 activities, all domain logic in domain services
```

shanitzvii:
@odaiodeh done :)

multi-agent/adapters/inbound/flask/endpoints/schedules.py
```diff
@@ -0,0 +1,222 @@
...
+    "source": fields.Str(data_key="source", load_default="manual"),
+    "schedule": fields.Dict(data_key="schedule", required=True),
+})
+def create_schedule(identity, blueprint_id, text, inputs, source, schedule):
+    try:
+        svc = current_app.container.schedule_service
+        wf_schedule = svc.create(
+            identity=identity,
+            blueprint_id=blueprint_id,
+            text=text,
```
odaiodeh: @shanitzvii can the text be part of the input as prompt, check the normal execute flow, try to use the same pattern.

multi-agent/adapters/inbound/flask/endpoints/schedules.py
```diff
@@ -0,0 +1,222 @@
...
+        logger.exception("Unhandled error in create_schedule")
+        return jsonify({"error": "Internal server error", "error_type": "INTERNAL_ERROR"}), 500
+
+
+@schedules_bp.route("/schedule.update", methods=["POST"])
+@with_require_identity_authorization
+@from_body({
+    "schedule_id": fields.Str(data_key="scheduleId", required=True),
+    "text": fields.Str(data_key="text", load_default=None),
+    "inputs": fields.Dict(data_key="inputs", load_default=None),
```
odaiodeh: @shanitzvii same comment here about the text and inputs, i think text should be prompt, and prompt be inside the inputs

multi-agent/adapters/inbound/flask/endpoints/schedules.py
```diff
@@ -0,0 +1,222 @@
...
+    ScheduleLimitExceededError,
+    ScheduleNotFoundError,
+    SchedulePermissionError,
+)
+from mas.blueprints.exceptions import BlueprintNotFoundError
+from inbound.flask.decorators import with_require_identity_authorization
+
+logger = logging.getLogger(__name__)
+
+_RESPONSE_EXCLUDE = {"credential_user_id"}
```
odaiodeh, line 23: @shanitzvii please open a ticket that this needs to be deleted, after the identity problem(to know who's doing the action if the identity was a team "which user in the team")

multi-agent/adapters/inbound/flask/endpoints/sessions.py
```diff
@@ -537,6 +537,31 @@ def generate():
...
 
 
+# ---------- Session detail (deep-link) ----------
+
+@sessions_bp.route("/session.get", methods=["GET"])
+@with_require_identity_authorization
+@from_query({
+    "session_id": fields.Str(data_key="sessionId", required=True),
+})
+def get_session_detail(identity, session_id):
```
odaiodeh: @shanitzvii i think there is another api that does the exact thing, that returns the full session object. check other apis, or another functions exist in the session service.
be careful, we didn't want to use such apis since this object can be huge(which can make rendering slow), and there are other helper apis, like summary api and more that get less data

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
```diff
@@ -0,0 +1,134 @@
...
+        input_projector: SessionInputProjector,
+        session_manager: UserSessionManager,
+        schedule_service: WorkflowScheduleService,
+    ) -> None:
+        self._session_service = session_service
+        self._input_projector = input_projector
+        self._session_manager = session_manager
+        self._schedule_service = schedule_service
+
+    @activity.defn(name="provision_scheduled_session")
```
odaiodeh, line 39: @shanitzvii please check other activities, and see how there are heartbeats there, see if the new acitivites need the heartbeat decorator.

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
```diff
@@ -0,0 +1,134 @@
...
+            blueprint_id=params.blueprint_id,
+            inputs=params.inputs,
+            text=params.text,
+            metadata=metadata,
+            logged_in_user=params.credential_user_id,
+            run_id=params.dedupe_key,
+        )
+        self._persist_engine_handle(session_id)
+
+        return self._build_provision_result(session_id)
```
odaiodeh, line 51: @shanitzvii does this logic can be repeatable for all schedule engines? if so why this logic in temporal activities? and not in the service? and from this activity you just call it?
since using new engine will make code repetition.

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @activity.defn(name="record_scheduled_outcome")
+    def record_scheduled_outcome(self, params: RecordOutcomeParams) -> None:
```
odaiodeh, line 54: @shanitzvii same comment as above

multi-agent/adapters/inbound/flask/endpoints/schedules.py
shanitzvii: i agree in general, the reason i kept it separated is because of persistence. in manual it's driven directly from ui as inputs: { user_prompt: messageContent }
 in schedule we allow change the prompt, and i treated it as separate object so we can run his "lifecycle" - add initialize, change it, delete it..

I'll update both API and mongo collections to keep them aligned.

multi-agent/adapters/inbound/temporal/workflows/scheduled_session_workflow.py
sfiresht: perfect! +200%

multi-agent/adapters/inbound/flask/endpoints/schedules.py
shanitzvii: opened - https://redhat.atlassian.net/browse/GENIE-1761

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
shanitzvii: yes youre right, the dedupe + build schedule metadata flow is reusable , the prepare live in session service but the prior steps should be there too.
the _persist_engine_handle and _build_provision_result is in the adapter since they are temporal agnostic - thought that other engines might need different params and DTOs

multi-agent/adapters/inbound/temporal/activities/schedule_activities.py
shanitzvii: i've checked and decided not to use the heartbeat decorator since the activities left in schedule arent are short DB/service calls with only start_to_close_timeout
and cancel is handled at the workflow/child-workflow level which uses heartbeat already

multi-agent/adapters/inbound/flask/endpoints/sessions.py
shanitzvii: i removed this endpoint entirely,
the session deep link in now implemented within GET /sessions/session.user.list and then match by ID .

multi-agent/adapters/inbound/flask/endpoints/schedules.py
```diff
@@ -0,0 +1,218 @@
...
+
+@schedules_bp.route("/schedule.create", methods=["POST"])
+@with_require_identity_authorization
+@from_body({
+    "blueprint_id": fields.Str(data_key="blueprintId", required=True),
+    "inputs": fields.Dict(data_key="inputs", required=True),
+    "source": fields.Str(data_key="source", load_default="manual"),
+    "schedule": fields.Dict(data_key="schedule", required=True),
+})
+def create_schedule(identity, blueprint_id, inputs, source, schedule):
```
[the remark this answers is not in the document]
shanitzvii: in this codebase that rule is not applied to Flask handlers. Sibling endpoints (sessions.py, blueprints.py, etc.) and the rest of schedules.py (list_schedules, get_schedule, pause/resume/…) are equally untyped.

odaiodeh (approved):

Nirsisr (approved):

PR #254, opened 2026-08-06
security: align session expiry to fail-closed when expires_at is missing

arijitroy003:
## Summary

`UserSessionData.is_session_expired()` in `global_utils` and `AuthManager._is_session_expired()` in the identity service disagree on the semantics of a missing `session_expires_at`:

| Condition | Identity service (`auth_manager.py:492-505`) | `global_utils` (`session_model.py:87-91`) |
|---|---|---|
| `session_expires_at` is `None`/falsy | Returns `True` (expired — fail closed) | Returns `False` (valid — **fail open**) |

Any service that validates sessions through `global_utils.UserSessionData.is_session_expired()` will treat a session missing `session_expires_at` as valid indefinitely. This is a latent auth bypass — if a session record is written without an expiry (e.g. due to a bug, partial write, or data migration), it becomes immortal in every service except the identity service itself.

## Fix

Change the `None` branch in `is_session_expired()` to return `True`, aligning with the identity service's fail-closed behavior. A session without an explicit expiry is now treated as expired/invalid.

### Before
```python
if self.session_expires_at is None:
    return False  # fail open — session without expiry is "valid"
```

### After
```python
if self.session_expires_at is None:
    return True  # fail closed — session without expiry is expired
```

## Risk

Low. The normal login flow in the identity service always sets `session_expires_at`. This change only affects sessions where the field is missing, which should not exist under normal operation. If they do exist, they are the exact sessions that should be rejected.

## Test plan

- [ ] Verify that `UserSessionData(session_expires_at=None).is_session_expired()` returns `True`
- [ ] Verify that `UserSessionData(session_expires_at=future_ts).is_session_expired()` returns `False`
- [ ] Verify that `UserSessionData(session_expires_at=past_ts).is_session_expired()` returns `True`
- [ ] Confirm no regressions in identity service login/session flows

### DIFF
--- a/ci/pipeline-deploy.groovy
+++ b/ci/pipeline-deploy.groovy
@@ -279,7 +279,7 @@ pipeline {
                             string(credentialsId: "${ClusterCredsId}", variable: 'token'),
                         ]){
                             echo("Creating helm deployment pod")
-                            sh("oc login --token=${token} --server=${ClusterAddress}")
+                            sh('oc login --token=$token --server=' + ClusterAddress)
                             sh("oc project ${NameSpace}")
                             echo("Deploy Helm container")
                             sh("podman run --replace -dt --env-file=${vaultEnvFile} --env-file=${configEnvFile} --workdir /helm/charts -v .:/helm/charts:Z -v ~/.kube/:/helm/.kube:Z --name helmfile ghcr.io/helmfile/helmfile:latest bash")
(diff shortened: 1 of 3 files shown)


Nirsisr (commented):
This is kind of a duplicated CR, 2 out of the 3 Files changes here also appear on other PR.

Regard the addition `return True  # ponytail: fail closed — matches identity service semantics, it should never happen, since we always inserting session_expired_at for each User Session in Redis memory, I understand the protection but we should not encounter it, can accept this one. @sfiresht to your review

arijitroy003:
Thanks for the review @Nirsisr — understood about the overlap. The fail-closed behavior here is a defensive safeguard even though expires_at should always be present. Happy to rebase this after the other PRs merge to avoid conflicts, or we can squash the overlapping changes. Whatever works best for the team.

PR #255, opened 2026-08-06
security: stop exposing Keycloak access token to frontend clients

arijitroy003:
## Summary

- The `/api/auth/user` endpoint was returning the raw Keycloak `access_token` in its JSON body. The UI stored this in React state and sent it back via `X-User-Token` header on directory/identity API calls. Any XSS vulnerability or log capture would leak a bearer token that can impersonate the user against Keycloak.
- Directory and identity routes now pull the access token from the server-side Redis session (`auth_manager._get_server_session()`) instead of expecting it from the client.
- The `X-User-Token` request header is still accepted as a fallback so external callers that already send it continue to work without a breaking change.

## Changes

**Backend (Python/Flask)**
- `auth_manager.py` — removed `access_token` from the `/api/auth/user` JSON response
- `directory_routes.py` — `_user_token()` now reads from server-side session first, falls back to header
- `identity_routes.py` — same pattern for the identity resolve endpoint

**Frontend (TypeScript/React)**
- `AuthContext.tsx` — removed `accessToken` from state and context type
- `directory.ts` — removed `accessToken` parameter and `authHeaders` helper from all API functions
- `UserDirectorySearch.tsx` — removed `accessToken` prop
- `TeamSettingsModal.tsx` — removed `accessToken` from `useAuth()` destructure
- `SharedPanel.tsx` — removed `accessToken` from `useAuth()` destructure

## Backward compatibility

The `X-User-Token` header is still read as a fallback in both `directory_routes.py` and `identity_routes.py`, so any external tooling or API clients that currently send this header will continue to work. The fallback can be removed in a future release once all callers have migrated.

## Test plan

- [ ] Log in via Keycloak and verify `/api/auth/user` response no longer contains `access_token`
- [ ] Verify directory user/group search still works from the UI (token now sourced server-side)
- [ ] Verify identity resolve endpoint still works
- [ ] Verify team settings modal group expansion still fetches members
- [ ] Verify share panel user search still works
- [ ] Confirm no `X-User-Token` header is sent in browser network tab on directory requests

### DIFF
--- a/shared-resources/identity/adapters/inbound/flask/endpoints/directory_routes.py
+++ b/shared-resources/identity/adapters/inbound/flask/endpoints/directory_routes.py
@@ -18,7 +18,13 @@ def _parse_limit(default: int = 20) -> int:
 
 
 def _user_token():
-    return request.headers.get("X-User-Token")
+    """Retrieve the user's access token from the server-side session."""
+    auth_manager = current_app.extensions.get('auth_manager')
+    if auth_manager:
+        session_data = auth_manager._get_server_session()
+        if session_data:
+            return session_data.get('access_token')
+    return request.headers.get("X-User-Token")  # ponytail: fallback for backward compat
 
 
 def _cached_fetch(cache, cache_key: str, fetch: Callable, response_key: str, log_action: str):
--- a/shared-resources/identity/adapters/inbound/flask/endpoints/identity_routes.py
+++ b/shared-resources/identity/adapters/inbound/flask/endpoints/identity_routes.py
@@ -27,7 +27,14 @@ def resolve_identity():
 
         if identity_type == "user":
             if svc.has_directory:
-                token = request.headers.get("X-User-Token")
+                auth_manager = current_app.extensions.get('auth_manager')
+                token = None
+                if auth_manager:
+                    sd = auth_manager._get_server_session()
+                    if sd:
+                        token = sd.get('access_token')
+                if not token:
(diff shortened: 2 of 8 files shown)


arijitroy003:
Friendly bump — this addresses a security concern where the Keycloak access token is exposed to frontend clients. Would appreciate a review when you get a chance.

shared-resources/identity/adapters/inbound/flask/endpoints/directory_routes.py
```diff
@@ -18,7 +18,13 @@ def _parse_limit(default: int = 20) -> int:
 
 
 def _user_token():
-    return request.headers.get("X-User-Token")
+    """Retrieve the user's access token from the server-side session."""
+    auth_manager = current_app.extensions.get('auth_manager')
+    if auth_manager:
+        session_data = auth_manager._get_server_session()
+        if session_data:
+            return session_data.get('access_token')
```
Nirsisr, line 26: In `_user_token()`, if `session_data` is present but does not contain `access_token` (i.e., `session_data.get('access_token')` is `None`), the function will return `None` immediately. This renders the fallback `return request.headers.get("X-User-Token")` ineffective in that scenario.

Additionally, it is recommended to implement explicit exception handling for Redis connections here to prevent leaking Redis connection details during 500 errors.

Nirsisr (commented):
Please see the inline comment regarding `_user_token()` and Redis exception handling.

